PR #9, opened 2023-09-03
compare custom encodig with and without brotli

Nuhvi:


js/benchmarks/README.md
```diff
@@ -37,11 +37,12 @@ Actually encoding ipv4 and ipv6 for what is mostly `A`, `AAAA`, and `TXT` record
  ┌--------------------┬------------┬----------┬------------┐
  | TYPE               | SIZE RATIO | COMPRESS | DECOMPRESS |
  |--------------------|------------|----------|------------|
- | json + brotli      | 0.97       | 1 ms     | 1 ms       |
- | bencode + lz4      | 1.01       | 0 ms     | 1 ms       |
- | csv + lz4          | 0.96       | 0 ms     | 1 ms       |
- | bencode + brotli   | 0.78       | 21 ms    | 0 ms       |
- | csv + brotli       | 0.77       | 10 ms    | 0 ms       |
- | custom + brotli    | 0.60       | 13 ms    | 5 ms       |
+ | json + brotli      | 0.96       | 1 ms     | 1 ms       |
```
dpc, line 40: milliseconds don't seem very useful. Would be good to increase iterations or use nanoseconds(?).

Nuhvi: Yes, I should make more accurate measurements as I implement this codec in Rust.

My goal with this is mostly to find a data-driven way to decide whether or not I should implement anything fancier than just JSON or Bencode encoding + compression.

Will update this soon.

Nuhvi: more in #10

PR #19, opened 2023-10-03
Add docker image.

dvc94ch:


Nuhvi:
@dvc94ch Thanks, but before I merge it is this to avoid installing Node and nvm etc. ? 
I wonder if that is the case we can add Nix support too, I am very new to it, but sounds lighter than Docker, maybe @dpc can chime in.

dvc94ch:
No, it's to deploy on kubernetes or other cloud infra. It's also useful during development when combining dozens of different systems.

dvc94ch:
And you still need to have node installed as it builds locally and only copies the artefacts to a docker image

PR #20, opened 2023-10-03
HTTP client in Rust

### LINKED ISSUE #13: HTTP client crate
As a part of the rewriting in Rust effort, we need a crate that uses an HTTP server as a proxy to the DHT, This allows us to use the JavaScript module `bittorrent-dht` while examining the best way to go on about writing a DHT client in Rust that is at least as good and reliable.

This client isn't a waste of time either, Eventually, after porting the DHT and server to Rust, we will still need this client to use in environments that do _not_ offer UDP sockets at all, like Replit or all browsers.

Tasks:
[x] - Update and simplify the server to be a transparent proxy.
[x] - Packets encoding in Rust (a document is a struct containing the collection of records).
[x] - Signing and verifying packets
[x] - Simple stateless HTTP client packets
[ ] - Get feedback from interested testers (including generating bindings for their languages)

Nuhvi:
closes #13 or at least opens the chance for serious review.


Nuhvi:
@dvc94ch You might want to take a look at this [commit](https://github.com/Nuhvi/pkarr/pull/20/commits/5d5f0c5fb4d3d958116c63e8b71e0237bf1fcec0) in particular as it creates a self referential struct that holds both the bytes of `v` as well as the decoded Packet, so the fallible initialization is _not_ delayed, `self.packet()` then is always a valid Packet, if either `from_bytes()` or `from_packet()` were OK.

At least that was my goal, I might have did something wrong.

dvc94ch:
No idea, never used safecell before. Looks interesting. I generally just avoid using self referential structs

Nuhvi:
It was generously suggested by @rklaehn so I trust it is good, I might be using it wrong, but will merge this anyways as long as there is no obvious bug in it.

PR #26, opened 2023-10-17
Remove openssl.

dvc94ch:


Nuhvi:
@dvc94ch Thanks, would you mind explaining the context of where did you find out you need this? just want to learn the reasoning.

dvc94ch:
It avoids linking to openssl and it lets you build for musl (make static builds).

dpc:
FWIW. In Rust projects we (my own projects and projects I contribute to) tend to avoid openssl as well. It complicates cross-compilation by a lot, and generally openssl has a CVE and requires patching every few months.

PR #36, opened 2023-11-22
Mainline support

### LINKED ISSUE #7: Rewrite in Rust
I was going to ask if there are any plans, but I've noticed that README says:

> Reimplement everything in Rust once the JS implementation is tested with some scale.

So I would be interested in using Rust implemenation and helping get it done.

Nuhvi:
closes #7


Nuhvi:
@rklaehn @dvc94ch I believe you will be affected by these changes if you update to `0.4.0` (not published yet), so would love to get your feedback.

PR #38, opened 2023-12-04
Api inconsistency in resolve/relay_get

### LINKED ISSUE #37: Api inconsistency in resolve/relay_get
Would be better if the relay_get returned a `Result<Option<SignedPacket>>`. An error should signal a networking error, no entry in the dht should return `None`.

It's also not clear why relay_put requires an owned packet. Since it gets serialized, a reference should be enough.

Nuhvi:
closes #37

### DIFF
--- a/pkarr/examples/async/relay-client.rs
+++ b/pkarr/examples/async/relay-client.rs
@@ -61,7 +61,7 @@ async fn main() -> Result<()> {
 
         let client = PkarrClient::new();
         client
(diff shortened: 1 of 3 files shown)


dvc94ch:
Thanks! Looks good

Nuhvi:
@dvc94ch Do you care about versioning purity at this point? don't want to whip a v2.0.0 just for this.

PR #51, opened 2024-02-18
feat: custom dht settings

### LINKED ISSUE #48: support configuring bootstrap nodes / bootstrapping the dht

Nuhvi:
closes #48 now that Mainline can work as a server that stores mutable values.


Nuhvi:
@dvc94ch please check if this work for your usecase

PR #56, opened 2024-02-28
fix: always include the pubkey in the inner bytes

Frando:
The inner bytes payload contained in a `SignedPacket` currently is different depending on how the `SignedPacket` was constructed. So the `AsRef<[u8]>` would returns a byte slice *with* the 32 byte pubkey at the beginning if the packet was created with `SignedPacket::from_packet()` and *without* the 32 byte pubkey if created with `SignedPacket::from_relay_response()`.

This PR fixes this and also exposes `as_bytes` / `from_bytes` methods which are useful for persisting the full signed packet to a database or cache.

### DIFF
--- a/pkarr/src/signed_packet.rs
+++ b/pkarr/src/signed_packet.rs
@@ -37,34 +37,47 @@ struct InnerParsed<'a> {
 }
 
 impl Inner {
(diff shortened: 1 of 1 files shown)


Nuhvi:
Wow can't believe it took so long for this bug to be discovered, thanks @Frando.

PR #57, opened 2024-03-01
fix: signed packet from bytes

Frando:
#56 had a mistake in the `from_bytes` method... Here's the fix, this time with the test to prevent mistakes like this again.

### DIFF
--- a/pkarr/src/signed_packet.rs
+++ b/pkarr/src/signed_packet.rs
@@ -50,7 +50,7 @@ impl Inner {
             Signature::from_bytes(bytes[32..96].try_into().expect("signature is 64 bytes"));
         let timestamp = u64::from_be_bytes(bytes[96..104].try_into().expect("seq is 8 bytes"));
 
-        let encoded_packet = &bytes.slice(72..);
+        let encoded_packet = &bytes.slice(104..);
 
         if verify_signature {
             public_key.verify(&signable(timestamp, encoded_packet), &signature)?;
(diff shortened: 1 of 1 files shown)


Nuhvi (approved):

Nuhvi:
@Frando Irrelevant but do you think the naming of the methods need changing to be idiomatic? like `try_from_bytes` instead? Similarly, is the `as_` methods naming appropriate because cloning Bytes is cheap? 

Sorry for the newbie questions, just want to leverage the V2 to most :)

Frando:
Yeah I am a bit unsure there usually myself... The general convention is:
`as_` -> gives you a reference to some inner data, always cheap
`to_` -> gives you an owned copy of some inner data, always cheap
`into_` -> gives you an owned value, can be expensive

So it should be `to_bytes` and not `as_bytes` because you get an owned `Bytes` and not a reference.

not sure when it's idiomatic to do `try_` prefixes to be honest. I think in std it's `try_from` to not conflict with `from` trait, and for regular methods it's `try_into_foo` if there's also a `into_foo` which would panic in case of error. E.g currently a lot of `try_with_capacity` etc methods are being added to `Vec` etc to fail with error and not panic if allocation fails.

Nuhvi:
Hmm I thought `try_` should be anything that returns Result. Will read more and update it in v2.

Frando:
Yea I've seen different things in different codebases. E.g. in databases or so where each operation involves IO it's not common to add `try_` prefixes because all methods would have `try_`. For fallible conversions I've seen both. Maybe there `try_` is more idiomatic, not sure.

PR #74, opened 2024-08-15
feat: update deps to only use rustls@0.23

dignifiedquire:


Nuhvi:
Thanks.
AFK today, will merge and release tomorrow.

PR #88, opened 2024-10-17
chore: add Nix flake and Nix dev shell using flakebox

dpc:
For Nix users like me, this allows bunch of interesting things:

`nix develop` will give a dev shell with Rust toolchain, ready to hack on the project.

Easy cross-compilation:

```
nix build .#aarch64-android.dev.pkarr-server && file result/bin/pkarr-server
result/bin/pkarr-server: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

```
> nix build .#riscv64-linux.dev.pkarr-server && file result/bin/pkarr-server
result/bin/pkarr-server: ELF 64-bit LSB pie executable, UCB RISC-V, RVC, double-float ABI, version 1 (SYSV), dynamically linked, interpreter /nix/store/7ac9zpgfw70544ca66dg1lsd4w8nyp0k-glibc-riscv64-unknown-linux-gnu-2.39-52/lib/ld-linux-riscv64-lp64d.so.1, for GNU/Linux 4.15.0, with debug_info, not stripped
```

After this lands any Nix user can seamlessly start `pkarr-server` with `nix run github:pubky/pkarr#pkarr-server`. Can be tested before that with:

```
> nix run github:dpc/pkarr/24-10-17-nixify#pkarr-server
2024-10-17T18:05:57.836178Z  INFO pkarr_server: Running as a resolver on UDP socket 0.0.0.0:6881
2024-10-17T18:05:57.836253Z  INFO pkarr_server::http_server: HTTP server listening on 0.0.0.0:6881
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 target/
 config.toml
+/result
+/.direnv
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,3 +5,10 @@ members = [
 
 # See: https://github.com/rust-lang/rust/issues/90148#issuecomment-949194352
 resolver = "2"
+
+[workspace.package]
+version = "0.5.0-alpha"
+
+[workspace.metadata.crane]
+name = "pkarr"
+
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,311 @@
+{
+  "nodes": {
+    "android-nixpkgs": {
+      "inputs": {
+        "devshell": "devshell",
+        "flake-utils": "flake-utils_3",
+        "nixpkgs": [
+          "flakebox",
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1727381935,
+        "narHash": "sha256-G2fOYRZM7bXK5eBb+GK3k/WmO+q5JA/GtFwSPc3kdc8=",
+        "owner": "tadfisher",
+        "repo": "android-nixpkgs",
+        "rev": "522d86121cbd413aff922c54f38106ecf8740107",
+        "type": "github"
+      },
+      "original": {
+        "owner": "tadfisher",
+        "repo": "android-nixpkgs",
+        "rev": "522d86121cbd413aff922c54f38106ecf8740107",
+        "type": "github"
+      }
+    },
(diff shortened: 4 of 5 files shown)


dpc:
BTW. Are you sure you don't want `pkarr-server` to be named `pkarrd`? :D

Nuhvi:
> BTW. Are you sure you don't want pkarr-server to be named pkarrd? :D

Pkarr server is a combination of a Relay (for browsers support) and a Resolver (Dht super node) to help lower latency on the native side.

They are not meant to be daemons, instead they are meant to be super nodes / gateways more like DNS resolvers than anything else.

You don't need them in native environment, and in browser, you are better off using globally popular relays (because they have better caches).

Of course, the Relay can still be used as local daemon for browser based applications like Electron, but that isn't what it is intended for.

Nuhvi:
@dpc would arguments pass naturally? so if I want to pass a path to config file would that work?

Also, I know almost nothing about Nix, so unless you want to add more things I can merge.

Finally, note that I am doing a lot of work on the `V3` branch, and I wonder if merging that to main will eventually affect you (I will pump the Cargo.toml version of course).

dpc:
> @dpc would arguments pass naturally? so if I want to pass a path to config file would that work?

An extra `--` might be needd so that `nix run ...` knows when the arguments to itself ended, but yeah, everything should work as expected.

```
> nix run github:dpc/pkarr/24-10-17-nixify#pkarr-server -- --help
Usage: pkarr-server [OPTIONS]

Options:
  -c, --config <CONFIG>                          Path to config file
  -t, --tracing-env-filter <TRACING_ENV_FILTER>  [tracing_subscriber::EnvFilter]
  -h, --help                                     Print help
```



> Finally, note that I am doing a lot of work on the `V3` branch, and I wonder if merging that to main will eventually affect you (I will pump the Cargo.toml version of course).

Do you plan to merge V3 into `main`? Anyway, it's all easily fixable, and from my PoV it's mostly about being able to bring dev version of Pkarr into a Nix config easily. The nixpkgs PR hopefully lands as well, so that would allow easily bringing the official releases.

Nuhvi:
Yes, I plant to merge V3, but not sure when, there is still couple of big features I need to add, and I don't want to merge what exists now without making sure these features won't require another breaking change.

PR #106, opened 2024-12-17
fix(pkarr): return cached packet in case the DHT resolved no more recent ones

### LINKED ISSUE #105: [Bug] resolve returns None on second call with `maximum_ttl=0`
Calling resolve 2 times in a row with `maximum_ttl(0)` results in `None` consistently on the second call. See example.

```
    #[test]
    fn ttl_0_test() {
        let uri = "7fmjpcuuzf54hw18bsgi3zihzyh4awseeuq5tmojefaezjbd64cy";
        let pubkey: PublicKey = uri.try_into().unwrap();
        let client = PkarrClientBuilder::default().maximum_ttl(0).build().unwrap();

        // First Call
        let value = client.resolve(&pubkey);
        match value.unwrap() {
            Some(val) => println!("Found1!"),
            None => println!("None1"),
        }
        sleep(Duration::from_secs(1));


        // Second Call
        let value = client.resolve(&pubkey);
        match value.unwrap() {
            Some(val) => println!("Found2!"),
            None => println!("None2"),
        }
        
    }
```

Result is always the same:

```
Found1!
None2
```

pkarr v2.2.0

Nuhvi:
closes #105

### DIFF
--- a/pkarr/src/client.rs
+++ b/pkarr/src/client.rs
@@ -286,7 +286,7 @@ impl PkarrClient {
 
             debug!(expires_in, "Have expired signed_packet in cache.");
         } else {
-            debug!("Cache mess");
+            debug!("Cache miss");
         }
 
         self.sender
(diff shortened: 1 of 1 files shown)


SeverinAlexB (approved):
I did some rough tests and it seems to work. Thx for the fix

PR #107, opened 2025-01-09
refactor: switch from simple-dns to hickory-proto

dignifiedquire:
Due to the segfaults we have seen with simple-dns, and to reduce the dependency load in `iroh` this switches to `hickory-proto`

Ref https://github.com/n0-computer/iroh/issues/2844

## Open Questions

- `packet` vs `message` naming, hickory and the RFC call the structure `Message`, but pkarr and simple-dns call them `Packet`
- optimize cloning of `SignedPacket`? see https://github.com/pubky/pkarr/pull/107#discussion_r1908940883
- should this again reexport hickory_proto, like it did `simple-dns` before?


pkarr/src/signed_packet.rs
```diff
@@ -1,66 +1,30 @@
...
-        })?)
-    }
-}
-
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, Clone)]
 /// Signed DNS packet
 pub struct SignedPacket {
-    inner: Inner,
+    message: Message,
```
rklaehn, line 32: A SignedPacket is now a wrapper around a hickory Message, which is a `Vec<Query>` and a bunch of `Vec<Record>`. The previous SignedPacket was basically free to clone and would be backed by at most 1 heap allocation. The new one is not cheap to clone and has possibly a large number of heap allocations.

rklaehn: See https://docs.rs/hickory-proto/latest/src/hickory_proto/op/message.rs.html#65-73

dignifiedquire: Yes, the alternative is to reparse the structure everytime you request details, like the previous version did for some parts.

dignifiedquire: It is not clear to me, what the usage patterns are in practice, and which ones should be optimized, so I went with the simplest and most typesafe one.

dignifiedquire: I don't think this actually gets mutated, so this could also be wrapped in an `Arc` under the hood, to optimize the cloning path again.

dignifiedquire: pushed a version that does the last variant

Nuhvi: > Yes, the alternative is to reparse the structure everytime you request details, like the previous version did for some parts.

No, the current version has the parsed [Packet](https://github.com/pubky/pkarr/blob/main/pkarr/src/signed_packet.rs#L27) already, no need to reparse, and that is memory efficient because simple-dns packet takes a reference of the bytes and read from them using offsets on demand.

> I don't think this actually gets mutated, so this could also be wrapped in an Arc under the hood, to optimize the cloning path again.

Yes it is not mutated after its creation, still, simple-dns Packet + Bytes, is more efficient than an Arc and a bunch of vectors.

But I can tolerate memory inefficiency if simple-dns can't be made robust.

pkarr/src/signed_packet.rs
```diff
@@ -190,25 +151,36 @@ impl SignedPacket {
...
+        bytes.extend(self.inner.message.to_vec().expect("valid message"));
+
+        bytes
     }
 
     /// Returns a slice of the serialized [SignedPacket] omitting the leading public_key,
     /// to be sent as a request/response body to or from [relays](https://github.com/Nuhvi/pkarr/blob/main/design/relays.md).
-    pub fn to_relay_payload(&self) -> Bytes {
-        self.inner.borrow_owner().slice(32..)
+    pub fn to_relay_payload(&self) -> Vec<u8> {
```
Nuhvi, line 205: Another example of allocation that isn't happening currently in SignedPacket, because currently the original packet is stored as is, which you can do even with hickory message, but then you would be doubling the amount of data, because simple-dns Packet doesn't copy the data, instead it just points to it.

Nuhvi:
Closing this after publishing `pkarr = "2.3.1"` with parsing segfaults fixed in simple-dns.

PR #108, opened 2025-01-12
chore(deps): bump simple-dns

Arqu:
`simple-dns` merged the fixes and published `0.9.2` which should hopefully cover all the segfault reports.


Nuhvi:
Published `pkarr = "2.3.1"` with the new version.

PR #116, opened 2025-02-15
fix(pkarr): remove rand and depend on getrandom only

Nuhvi:


pkarr/src/extra/endpoints/endpoint.rs
```diff
@@ -187,6 +184,17 @@ fn get_svcb<'a>(record: &'a ResourceRecord, get_https: bool) -> Option<&'a SVCB<
     }
 }
 
+/// Shuffles a slice randomly.
+fn shuffle(slice: &mut [&SVCB<'_>]) {
+    for i in 1..slice.len() {
+        let mut buf = [0u8; 8];
+        getrandom::getrandom(&mut buf).expect("getrandom failed");
```
dpc, line 191: Is this motivated with the trend to avoid chunky `rand`?

If `getrandom` is a always a syscall, than this thing might be kind of slowish (on a very micro level of course). And `rand_pos` will be biased. Not sure if it matter here at all.

Nuhvi: The actual motivation is that `rand 0.9` sucked for being chunky but also because `getrandom 0.3.1` sucks in wasm.

But then this is the only place where I use any of `rand` fanciness, so I figured I would do it myself instead.

If this became an issue, do what rand does, while not worrying about its drama.

dpc: Makes sense.

PR #119, opened 2025-02-17
chore: add config.toml to .gitignore

MiguelMedeiros:
Add `config.toml` to `.gitignore` to prevent committing local configuration files. This ensures that sensitive configuration data and personal settings remain local and are not accidentally pushed to the repository.

The example configuration file (`config.example.toml`) will remain tracked to serve as a template for users.

Changes:
- Added `relay/config.toml` to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target/
 /result
 /.direnv
+config.toml


Nuhvi:
@MiguelMedeiros It is a leftover, when it was useful, it shouldn't have been gitignored, and now it should just be deleted.

MiguelMedeiros:
The fix has already been made in this [commit](https://github.com/pubky/pkarr/pull/121/commits/0f202eb55d19c3779707e7ddbd497c3e0f676d3e), so the file is no longer needed.

PR #120, opened 2025-02-17
fix: prevent thread leak in LmdbCache stats reporting

MiguelMedeiros:
## Problem
While the pkarr relay server typically uses a single LmdbCache instance, the `LmdbCache` implementation (available in the `extra` module) had a background thread for logging cache statistics that never terminated when the cache was dropped. 

Each `LmdbCache::open()` call creates a new instance with its own stats thread. Although this isn't an issue in the relay server's single-cache usage, it could cause resource leaks in projects using pkarr as a library where multiple cache instances might be created and destroyed.

## Solution
Added proper thread lifecycle management using `AtomicBool` to ensure the stats thread is terminated when the cache is dropped:

- Added `running: Arc<AtomicBool>` field to control thread lifecycle
- Implemented `Drop` trait to stop the thread when cache is dropped
- Added test to verify thread cleanup

## Testing
Added `test_stats_thread_leak` that verifies:
- Stats thread starts when cache is created
- Thread properly terminates when cache is dropped
- No thread leaks occur

This change is fully backwards compatible as it only affects internal implementation details.

### DIFF
--- a/pkarr/src/extra/lmdb_cache.rs
+++ b/pkarr/src/extra/lmdb_cache.rs
@@ -7,6 +7,7 @@ use std::{
     path::Path,
     sync::{Arc, RwLock},
     time::Duration,
+    sync::atomic::{AtomicBool, Ordering},
 };
 
 use byteorder::BigEndian;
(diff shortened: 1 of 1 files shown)


Nuhvi:
@MiguelMedeiros Nice catch, but the right thing to do here is to remove that thread entirely, and possibly adding this logging feature to the relay directly.

Nuhvi:
Fixed at 2aa1c305804aa3eee1806bd270e7616559be7cef

PR #124, opened 2025-02-18
feat: add dedicated `/info` endpoint for relay statistics

MiguelMedeiros:
This PR introduces a dedicated `/info` endpoint that returns relay statistics in JSON format.

Changes:
- Add new `/info` endpoint that returns relay stats in JSON format
- Keep index endpoint focused on HTML representation
- Maintain same statistics data but with cleaner API design
- Add `serde_json = "1.0"` dependency for JSON serialization

This change provides:
- More RESTful API design
- Clearer separation of concerns between HTML and JSON responses
- Better discoverability of the JSON API
- Improved maintainability with dedicated endpoints for each format

The endpoint returns the same statistics as before:
- Relay version
- Cache statistics (size, capacity, utilization)
- DHT information (node port, firewall status, network size, confidence)

Added test for `/info` endpoint that verifies:
- Response status code
- JSON structure
- Field value ranges
- Optional confidence field handling

You can run the test with:
```bash
cargo test -p pkarr-relay handlers::tests::test_info_endpoint
```

To access the new JSON stats, visit: http://localhost:6881/info

Example response:
image


Nuhvi:
Why? who needs this? Any machine going to read this? Can we remove or change it later without breaking the code of someone who depended on it without our knowledge?

MiguelMedeiros:
>Why? who needs this? Any machine going to read this? 

I want to aggregate stats from multiple relays for meta-analysis. Comparing performance metrics across them. This JSON endpoint makes that easier. 

>Can we remove or change it later without breaking the code of someone who depended on it without our knowledge?

If we ever need to change the format, we can version it (e.g., `/info/v1` and `/info/v2`) to ensure backward compatibility.

Nuhvi:
I personally want relays API to stay as simple as described in the `relays.md`,  and not make things any more complex.

I am also not convinced of the use case, so I disapprove, but I don't own this repo.

MiguelMedeiros:
>I personally want relays API to stay as simple

I understand the need for simplicity. My end goal is to eventually create a scoring system for each relay, and having a JSON endpoint makes it easier to aggregate and compare stats across multiple relays for that meta-analysis. 

However, I appreciate your concerns about complexity and don’t want to create unnecessary stress or disruption. I'll close this PR for now. Thanks for your feedback.

PR #129, opened 2025-02-21
upgrade and re-export mainline, and remove unnecessary methods in wasm

Nuhvi:


Nuhvi:
Github CI seems to be broken, even older commits that were passing now fail.. so I will just pretend Github is not available today.

PR #142, opened 2025-05-16
fix(pkarr): merge dht relay streams

SHAcollision:
Early relay errors (rate limits, not-found) prematurely cancel ongoing DHT lookups. Can be replicated using the resolve example and using as relay a mock-up service that instantly returns an error. We've observed in practice this issue happening to Nexus when resolving user keys during stress tests.

**Problem:** when both DHT and relays are enabled, the previous `futures_lite::future::or` combinator would terminate the merged resolve stream as soon as all relays responded (even with errors), preventing any subsequent DHT results from surfacing.

**Fix:** replace the `or` combinator with `futures_util::stream::select` in `Client::merged_resolve_stream`, so that items from both streams are interleaved and each continues independently until both are exhausted. Additional dependencies introduced:  `futures_util`

TODO: we have a similar issue when publishing:
https://github.com/pubky/pkarr/blob/0f10cd813c600d039cc85e8867b698b42bdd6fac/pkarr/src/client.rs#L349-L354
If relays are down or misbehaving, the DHT publish is skipped.


SHAcollision:
The test `client::tests::concurrent_publish_different_with_cas::case_2_both_networks` became flaky. It seems to consistently fail on the github runner. It is succeeding locally, but if you retry enough, it also fails randomly.

```
──── STDERR:             pkarr client::tests::concurrent_publish_different_with_cas::case_2_both_networks

thread 'client::tests::concurrent_publish_different_with_cas::case_2_both_networks' panicked at pkarr/src/client/tests.rs:384:51:
called `Result::unwrap()` on an `Err` value: Concurrency(CasFailed)
```

My guess is by switching from `futures_lite::stream::or` to `futures_util::stream::select` we changed the polling order and lifetime of the two branches in a way that the next `publish(..., Some(cas))`  hits the “CAS failed” branch.

Unfortunately I'm not capable to understanding or debugging further this.

Nuhvi:
@SHAcollision Thanks... I will look into this today.

Nuhvi:
- Client::resolve(): stream select is now implemented locally with no dependencies.
- Client::publish(): publishing now is sequential, first to the DHT then to relays, although if publishing to the DHT is successful, publishing to relays is done in the background.

Nuhvi:
@SHAcollision 

Thee bad news is publishing is sequential; DHT first then relays and only in the background if the dht succeeded.

The good news is I think you won't have an issue with relays screwing the DHT, however it is up to you to write the regression test for your case, if at all. 

I will make a release if you think this works fine.

SHAcollision:
Thanks a lot @Nuhvi :rocket: 

IMO the fix looks solid. I've tested and both the resolve and publishing issues detailed above are solved.

I'm only worried by a side effect this PR might have introduced. I propagated this patch through our stack and re-run the stress test. The stress test is basic: signup users to the homeserver in a loop and write a single file by each new user, that's it. I've done this more than 100K times already. However, when using the `pkarr-client` from this branch, after 7 to 20 `signup` & `publish_homeserver` iterations the script dies with the following message:
```
thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```
I've run this script multiple times for over 10K iterations without seeing this. Now it is consistenly dying before iteration 20. This could likely be some weird interaction between this fix and the way we use the `pkarr-client` within the `pubky-client` that leads to deep recursion or a very large async‐state. On Monday, I will try to find whether I can create an MRE of this issue using solely `pkarr` crate.

Nuhvi:
@SHAcollision Yeah I wouldn't merge or release this yet.

That being said, as I mentioned before, there is no way a client can work for all edge cases and stressful tasks. I still think for bulk republishing, you are better off using the SignedPacket and Mainline, I definitely think a republisher shouldn't be using relays network at all.

Note that `signup` and `publish_homeserver` are supposed to happen very rarely on client side, not sure why are you stress testing that at all.

We need to make the client more robust, but be careful not to waste days trying to fix for a use case that shouldn't exist at all.

SHAcollision:
> but be careful not to waste days trying to fix for a use case that shouldn't exist at all.

You are right.

We are overally finding the limits on how much "user activity" can we create to the homeserver and Nexus sync. It's working OK to poorly, many events Nexus can't sync because it cannot find the user's homeserver. So I scoped it down to simply try to signup -> write a `profile.json` -> check whether the user even get indexed. I'm finding that only about 7000 out of 10000 get indexed.

We have a republisher in the homeserver that only uses `dht` . But it seems important that the clients / nexus can use the relays because these records seem to only have a 80-95% chance to be found by Nexus directly on DHT (if both the client and nexus are `.no_relays()`). It is also an edge case as you say that a single client is trying to signup 1000 users. Maybe I should try again by cycling the client ( e.g. 1 client per user). But certainly Nexus will be a single client trying to fetch LOTs of data (resolving pkarr only! not publishing so probably no prob).

Edit: Ummm using a single `pubky-client` per user I still hit this before the 5-10 user.
```
thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

Nuhvi:
@SHAcollision I said this before and I will repeat it again, testing Pkarr as a part of testing Nexus/Homeservers is like testing DNS as a part of testing a web server. Either you take Pkarr for granted or you don't, if you do (after separate testing, since it is really a separate thing), then you should design Pubky client/homeserver and nexus to skip/mock pkarr altogether.

Feel free to signup 10000000000 users to a homeserver, but trying to do that + publishing that many users to a small DHT and couple of relays ... is always going to give you grief and waste your time.

pkarr/src/client.rs
```diff
@@ -484,7 +489,7 @@ impl Client {
         #[cfg(all(dht, relays))]
         Box::pin(match (dht_stream, relays_stream) {
             (Some(s), None) | (None, Some(s)) => s,
-            (Some(a), Some(b)) => Box::pin(futures_lite::stream::or(a, b)),
+            (Some(a), Some(b)) => Box::pin(select_stream(a, b)),
             (None, None) => unreachable!("should not create a client with no network"),
         })
```
SHAcollision: Is this `Box::pin(match {...})` needed?
```rust
        match (dht_stream, relays_stream) {
            (Some(s), None) | (None, Some(s)) => s,
            (Some(a), Some(b)) => Box::pin(select_stream(a, b)),
            (None, None) => unreachable!("should not create a client with no network"),
        }
 ```

**Comment by Nuhvi** [comment]
> Client::resolve(): stream select is now implemented locally with no dependencies.
Client::publish(): publishing now is sequential, first to the DHT then to relays, although if publishing to the DHT is successful, publishing to relays is done in the background.

Turned out the concurrent publish wasn't the problem at all, I had a bad test. The squash message should be something like this now:

- Client::resolve(): stream select is now implemented locally with no dependencies.
- Client::publish(): while publishing to both networks (dht/relays), wait for both queries to be complete.
- Client::publish(): while publishing to both networks (dht/relays), if the first finished query returned a success, ignore the CAS errors from the later done query, as it is expected due to outdated CAS argument, but both networks would be successfully updated to the last timestamp regardless.

**Comment by SHAcollision** [comment]
> thread 'tokio-runtime-worker' has overflowed its stack
> fatal runtime error: stack overflow
> Aborted (core dumped)


This is an MRE for it. It's pretty inconsistent: often crashing before iteration 20, I've seen it crashing even in the first 5 iters, but I've seen it also go further than 5000 iters and not crash!! I've tried on 2 different machines and seen it happen in both. Never seen before this PR.
```rust
use pkarr::{dns::rdata::SVCB, Client, ClientBuilder, Keypair, SignedPacket};

#[tokio::main]
async fn main() -> () {
    let host: pkarr::dns::Name = "example.com".try_into().unwrap();

    for i in 0.. {
        // we can build a new client per iteration or not. Does not
        // affect the outcome, it crashes regardless.
        let client: Client = ClientBuilder::default().build().unwrap();

        // 1) do the resolve_most_recent
        let existing = client
            .resolve_most_recent(&Keypair::random().public_key())
            .await;

        // 2) build a new `_pubky` packet
        let kp = Keypair::random();
        let svcb = SVCB::new(0, host.clone());
        let pkt = SignedPacket::builder()
            .https("_pubky".try_into().unwrap(), svcb, 60 * 60)
            .sign(&kp)
            .unwrap();

        // 3) publish with CAS
        let cas = existing.map(|p| p.timestamp());
        let _ = client.publish(&pkt, cas).await;

        println!("iteration {}", i);
    }
}
```
Runs as:
```
   Compiling pkarr v3.7.1 (/home/crypt/pubky/pubky-stack/pkarr/pkarr)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 6.51s
     Running `target/debug/examples/publish`
iteration 0
iteration 1
iteration 2
iteration 3
iteration 4
iteration 5
iteration 6
iteration 7
iteration 8
iteration 9
iteration 10
iteration 11
iteration 12
iteration 13
iteration 14
iteration 15
iteration 16
iteration 17
iteration 18
iteration 19
iteration 20
iteration 21

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

The previous MRE is a best effort to do functionally the same as this `pubky-client` usage:
```
use pubky::{Client, Keypair};

#[tokio::main]
async fn main() -> () {
    let homeserver = Keypair::random().public_key();

    for i in 0..10_000 {
        let client = Client::builder().build().unwrap();
        let keypair = Keypair::random();
        let _ = client.republish_homeserver(&keypair, &homeserver).await;
        println!("Published homeserver: {}", i);
    }
}
```
```
Published homeserver: 0
Published homeserver: 1
Published homeserver: 2
Published homeserver: 3
Published homeserver: 4
Published homeserver: 5
Published homeserver: 6
Published homeserver: 7
Published homeserver: 8
Published homeserver: 9
Published homeserver: 10
Published homeserver: 11
Published homeserver: 12
Published homeserver: 13
Published homeserver: 14
Published homeserver: 15

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```

This flaky tokio issue is introduced by the fix to the `pkarr_client.publish()` in this PR.

pkarr/src/client.rs
```diff
@@ -325,19 +323,24 @@ impl Client {
...
         #[cfg(dht)]
-        let dht_future = self.dht().map(|node| async {
-            node.as_async()
-                .put_mutable(signed_packet.into(), cas.map(|t| t.as_u64() as i64))
-                .await
-                .map(|_| Ok(()))?
-        });
+        let dht_future = {
+            let signed_packet = signed_packet.clone();
+            self.dht().clone().map(|node| async move {
```
SHAcollision:
```suggestion
            self.dht().map(|node| async move {
```

Nuhvi:
@SHAcollision thanks, will look into it in an hour after work.

Nuhvi:
@SHAcollision 

What is this? like, why do you create 10K clients of anything? what does that simulate?

Also, why are you doing this with actual real network and not a testnet? what am I missing here?

```rust
use pubky::{Client, Keypair};

#[tokio::main]
async fn main() -> () {
    let homeserver = Keypair::random().public_key();

    for i in 0..10_000 {
        let client = Client::builder().build().unwrap();
        let keypair = Keypair::random();
        let _ = client.republish_homeserver(&keypair, &homeserver).await;
        println!("Published homeserver: {}", i);
    }
}
```

Nuhvi:
For the record, the only way I can replicate this is when I enable relays (with or without dht) AND reduce the request timeout significantly.

This tells me that http timeouts are causing this leak.

Interestingly, this can't be replicated in a local test network, it only works using default live relays.

Here is an example of that: 

```rust
use std::time::Duration;

use pkarr::{dns::rdata::SVCB, Client, ClientBuilder, Keypair, SignedPacket};

#[tokio::main]
async fn main() -> () {
    tracing_subscriber::fmt()
        .with_env_filter("pkarr=debug")
        .init();

    let host: pkarr::dns::Name = "example.com".try_into().unwrap();

    for i in 0.. {
        // we can build a new client per iteration or not. Does not
        // affect the outcome, it crashes regardless.
        let client: Client = ClientBuilder::default()
            .no_dht()
            .request_timeout(Duration::from_millis(100))
            .build()
            .unwrap();

        // 1) do the resolve_most_recent
        let existing = client
            .resolve_most_recent(&Keypair::random().public_key())
            .await;

        // 2) build a new `_pubky` packet
        let kp = Keypair::random();
        let svcb = SVCB::new(0, host.clone());
        let pkt = SignedPacket::builder()
            .https("_pubky".try_into().unwrap(), svcb, 60 * 60)
            .sign(&kp)
            .unwrap();

        // 3) publish with CAS
        let cas = existing.map(|p| p.timestamp());
        let _ = client.publish(&pkt, cas).await.unwrap();

        println!("\niteration {}", i);
    }
}
```

The test that doesn't fail:

```rust
#[rstest]
#[case::dht(Networks::Dht)]
#[case::both_networks(Networks::Both)]
#[cfg_attr(feature = "relays", case::relays(Networks::Relays))]
#[tokio::test]
async fn regression_tokio_stack_overflow(#[case] networks: Networks) {
    let testnet = mainline::Testnet::new_async(5).await.unwrap();
    let relay = Relay::run_test(&testnet).await.unwrap();

    let host: crate::dns::Name = "example.com".try_into().unwrap();
    let svcb = SVCB::new(0, host);
    let signed_packet_builder =
        SignedPacket::builder().https("_pubky".try_into().unwrap(), svcb.clone(), 60 * 60);

    let client_builder = builder(&relay, &testnet, networks);

    for i in 0..300 {
        // we can build a new client per iteration or not. Does not
        // affect the outcome, it crashes regardless.
        let client: Client = client_builder.build().unwrap();

        // 1) do the resolve_most_recent
        let existing = client
            .resolve_most_recent(&Keypair::random().public_key())
            .await;

        // 2) build a new `_pubky` packet
        let kp = Keypair::random();
        let pkt = signed_packet_builder.clone().sign(&kp).unwrap();

        // 3) publish with CAS
        let cas = existing.map(|p| p.timestamp());
        let _ = client.publish(&pkt, cas).await;

        println!("iteration {}", i);
    }
}
```

SHAcollision:
> @SHAcollision
> 
> What is this? like, why do you create 10K clients of anything? what does that simulate?
> 
> Also, why are you doing this with actual real network and not a testnet? what am I missing here?
> 
> ```rust
> use pubky::{Client, Keypair};
> 
> #[tokio::main]
> async fn main() -> () {
>     let homeserver = Keypair::random().public_key();
> 
>     for i in 0..10_000 {
>         let client = Client::builder().build().unwrap();
>         let keypair = Keypair::random();
>         let _ = client.republish_homeserver(&keypair, &homeserver).await;
>         println!("Published homeserver: {}", i);
>     }
> }
> ```

Sorry, clicked "close pull request" instead of "comment". Scary how close those two buttons are.

I made this MRE only for reference here. Set a big loop 10K as example, I'm not creating 10K clients anywhere.

Nuhvi:
@SHAcollision I found exactly [where](https://github.com/pubky/pkarr/blob/main/pkarr/src/client/relays.rs#L266) the stack overflow is happening, but it is not easy to figure out why it is happening, so this will take some time.

I pushed a commit with the regression test that I am trying to pass.

SHAcollision:
> @SHAcollision I found exactly [where](https://github.com/pubky/pkarr/blob/main/pkarr/src/client/relays.rs#L266) the stack overflow is happening, but it is not easy to figure out why it is happening, so this will take some time.

Indeed, it seems to be something here. I applied this patch (that very likely breaks something else!!!) and the test you created now passes.
```rust
--- a/src/client/relays.rs
+++ b/src/client/relays.rs
@@ impl InflightPublishRequests {
-    /// Returns true if request is done.
     fn add_error(
         &mut self,
         public_key: &PublicKey,
         error: PublishError,
     ) -> Result<bool, PublishError> {
         let mut inflight = self.requests.lock().expect("InflightPublishRequests lock");
 
         if let Some(request) = inflight.get_mut(public_key) {
             let majority = self.relays_count / 2 + self.relays_count % 2;
 
-            // Add error, and return early error if necessary.
-            {
-                let count = request.errors.get(&error).unwrap_or(&0) + 1;
-
-                if count >= majority
-                    && matches!(
-                        error,
-                        PublishError::Concurrency(ConcurrencyError::NotMostRecent)
-                    ) | matches!(
-                        error,
-                        PublishError::Concurrency(ConcurrencyError::CasFailed)
-                    )
-                {
-                    inflight.remove(public_key);
-
-                    return Err(error);
-                }
-
-                request.errors.insert(error, count);
-            }
+            // Count this error; if we’ve now seen a majority, fail immediately.
+            let count = request.errors.get(&error).unwrap_or(&0) + 1;
+            request.errors.insert(error.clone(), count);
+            if count >= majority {
+                inflight.remove(public_key);
+                return Err(error);
+            }
 
             // If we’ve heard from *all* relays, then we’re done:
             if self.done(request) {
                 let request = inflight.remove(public_key).expect("infallible");
 
                 if request.success_count >= majority {
                     Ok(true)
                 } else {
                     let most_common_error = request
                         .errors
                         .into_iter()
                         .max_by_key(|&(_, count)| count)
                         .map(|(err, _)| err)
                         .expect("infallible");
 
                     Err(most_common_error)
                 }
             } else {
                 Ok(false)
             }
         } else {
             Ok(true)
         }
     }
```

Nuhvi:
@SHAcollision yes I did the same, but on one hand this code existed for a reason, and on the other hand, I NEED to understand how is calling `HashMap::get` causing a stack overflow.

Nuhvi:
@SHAcollision done.

Nuhvi (approved):

SHAcollision:
Thanks a lot @Nuhvi been playing with it and propagated it through the stack and seems to work just fine :ok_hand: Let's merge it.

Issue #42, opened 2024-01-26
Consider storing historical data


thereisnogabe:
Add a query parameter to the GET API to query by sequence number.

Allow relays to optionally store historical data, which may be useful.

Nuhvi:
@decentralgabe sorry missed this, have you implemented this somewhere and suggesting to add it to the relay API design?

I don't have a strong opinion against it except for managing expectations. Would like to see your implementation if any.

Nuhvi:
I wouldn't want to promise any persistence or even a large cache. Relays were meant for helping browsers not much more.

thereisnogabe:
> have you implemented this somewhere and suggesting to add it to the relay API design

impl is in progress, but it is in our spec under [Historical Resolution](https://did-dht.com/#historical-resolution)

we pretty much return a sorted list of seq numbers when you do a GET for an identifier, and then you can add a query param for a specific seq number

> I wouldn't want to promise any persistence or even a large cache. Relays were meant for helping browsers not much more.

Agree, don't think it should be a requirement but optional for relays that wish to do this.

Nuhvi:
@decentralgabe storing historical data is a good idea for many use cases and could even serve as soft key rotation mechanism if enough watch towers are around. But I want to keep an eye first on your efforts for rate limiting and see if it works, scales, and endures better than ION.

Keeping this issue open for that.

PR #145, opened 2025-06-11
feat: mvp for persistent cache

evd0kim:
I propose to evaluate an optional persistent cache for pkarr which may:

* increase reliability of the service when it works along with other components/microservices
* partially address the issue https://github.com/pubky/pkarr/issues/42 for which I could add a method/endpoint for getting currently cached records from the service. An external indexer could track changes in pkarr records. 

Let me know your thoughts on this @MiguelMedeiros @SHAcollision @SeverinAlexB.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1414,7 +1414,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc2f4eb4bc735547cfed7c0a4922cbd04a4655978c09b54f1f7b228750664c34"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Nuhvi:
1. There is already a persistent cache optional for anyone who wants to use it https://github.com/pubky/pkarr/blob/main/pkarr/src/extra/lmdb_cache.rs, and indeed it is used in the relays to persist packets until the cache is full.
2. If you particularly dislike LMDB, you can build any other persistent cache primitive using the `Cache` trait, and pass it to the client, but no need to add it to this crate.
3. Serialization of everything in Pkarr is tightly defined, there is no reason to use JSON at all, and we try to reduce the amount of dependencies in Pkarr as much as possible.
4. A Persistent cache doesn't solve #42, simply because neither the `Client` nor the [Relays](https://github.com/Pubky/pkarr/blob/main/design/relays.md) have APIs to query for older versions of a packet. I also think we should never add such an API, since we should never make the promise of data persistence at any layer, Relays don't persist data beyond their limited cache size, nor is the DHT. Someone may build a custom relay that act as an archive, they are free to do so, but it shouldn't be endorsed by Pkarr as it serves nothing but confusing users to think that Pkarr offers some sort of Data Availability guarantees that it definitely does not, in fact the very philosophy of Pkarr is that not even Blockchains can guarantee DA for prunable data, otherwise, blockchains would have been superior to Pkarr as it would have offered key rotation and revocation, but that promise can't be delivered upon.

evd0kim:
Thanks for the feedback @Nuhvi. Everything makes sense to me now.

PR #149, opened 2025-06-17
fix: devShells now find openssl

TheButlah:
Super convenient that this library already has the flake for a devshell! However on NixOS its missing libssl. I added it along with pkg-config.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,6 +17,9 @@
     flake-utils.lib.eachDefaultSystem (
       system:
       let
+        pkgs = import nixpkgs {
+          inherit system;
+        };
         flakeboxLib = flakebox.lib.${system} {
           config = {
             github.ci.buildOutputs = [ ];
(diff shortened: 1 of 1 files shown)


Nuhvi:
I don't know much about Nix, and too lazy to try and learn all of that, but what I know is that we shouldn't depend on libssl, so if we do, we should find out where and replace it with `rustls` instead, and if the dependency doesn't support replacing `libssl` with `rustls` then we should consider replacing the dependency itself.

Nuhvi:
Maybe we should configure some dependency with `no-default-features` or something... needs some investigation.

TheButlah:
(I second the opinion about libssl).
It gets pulled in by request's [default-tls](https://docs.rs/crate/reqwest/latest/features#default-tls) feature. This feature enables rustls which enables aws-lc-sys as its default provider/backend. You have to configure reqwest to properly configure rustls to not use a default backend, by adding the `rustls-tls-no-provider` to reqwest.

Nuhvi:
closing in favor of #152

PR #152, opened 2025-06-17
remove openssl

Nuhvi:
@SHAcollision could you verify that this solves all dependencies on openssl for docker, and maybe update the docker file accordingly?

@TheButlah could you check if this solves the Nix flake issue?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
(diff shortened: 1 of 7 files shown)


SHAcollision:
> @SHAcollision could you verify that this solves all dependencies on openssl for docker, and maybe update the docker file accordingly?

It does :ok_hand: Opened PR https://github.com/pubky/pkarr/pull/154/files

SHAcollision (approved):

PR #153, opened 2025-06-17
chore: JS bindings

### LINKED ISSUE #146: Update JS package to current Pkarr implementation
Currently the JS package of `pkarr` is based on legacy JS implementation of Pkarr https://www.npmjs.com/package/pkarr last published over a year ago.

To use latest pkarr as defined in this project in browser, downstream projects such as `pubky-client` build a WASM package and could potentially expose methods to `publish()` and `resolve()` pkarr `SignedPackets`. However, using a `pubky-client` wrapping pkarr for relay interaction and packet parsing is excessive for lightweight use cases.

The proposal is to build a minimalistic `wasm-pack` JS package of the Pkarr client capable of performing the functions displayed in https://app.pkarr.org and to be used as a replacement of the current `npmjs` package on all environments.

tipogi:
Implements comprehensive WebAssembly bindings for Pkarr packet publishing/resolving with enhanced DNS support and error handling

## WASM API
- **Client** - Publish/resolve pkarr packets via Pubky relays with async support and custom configuration
- **Keypair** - Ed25519 keypair generation and management with validation
- **SignedPacket** - Fluent API for building pkarr packets with DNS record types (TXT, A, AAAA, CNAME, HTTPS, SVCB, NS)
- **Error Handling** - Centralized ClientError system with input validation, DNS names, IP addresses, timeouts, packet size limits
- **TypeScript Definitions** - Full type support for all WASM exports


pkarr/src/wasm/builder.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /// Add a TXT record to the packet
+    ///
+    /// # Arguments
+    /// * `name` - The domain name (e.g., "example" or "subdomain.example")
+    /// * `text` - The text content
+    /// * `ttl` - Time to live in seconds
+    #[wasm_bindgen(js_name = "addTxtRecord")]
+    pub fn add_txt_record(&mut self, name: &str, text: &str, ttl: u32) -> Result<(), JsValue> {
```
SHAcollision: Near identical 7 times the same function for each record type. There is maybe a possibility to reduce boilerplate in some way. For example a macro kind of like this? Might not be general enough, not sure:
```rust
macro_rules! impl_record {
  ($fn_name:ident, $inner_fn:ident, $js_name:expr, $rdata:ty) => {
    #[wasm_bindgen(js_name = $js_name)]
    pub fn $fn_name(&mut self, name: &str, /* … */) -> Result<(), JsValue> {
      self.validate_name(name)?;
      let domain_name = Name::new_unchecked(name);
      self.inner = self.inner.clone().$inner_fn(domain_name, /* … */);
      Ok(())
    }
  };
}
```

pkarr/pkg/test/all.js
```diff
@@ -0,0 +1,230 @@
...
+    console.log('=' .repeat(80));
+    console.log('🚀 Starting complete test execution...\n');
+    
+    const startTime = Date.now();
+    const results = {
+        unit: { status: 'pending', duration: 0, error: null },
+        integration: { status: 'pending', duration: 0, error: null },
+        performance: { status: 'pending', duration: 0, error: null },
+        edgeCases: { status: 'pending', duration: 0, error: null }
+    };
```
SHAcollision: Could we reduce `all.js` entirely by automatically discovering and running the other four suites via a simple script, rather than inlining their runners?

pkarr/pkg/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "pkarr",
+  "collaborators": [
+    "Nuh <nuh@nuh.dev>"
+  ],
+  "description": "Public-Key Addressable Resource Records (Pkarr); publish and resolve DNS records over Mainline DHT",
+  "version": "3.7.1",
```
SHAcollision: Not current, I guess a rebase is needed?

pkarr/pkg/test/all.js
SHAcollision: 100% AI suggested and untested
```js
/**
 * pkg/test/all.js
 *
 * Discovers and runs every test file in this directory.
 */

const fs = require('fs');
const path = require('path');

async function runSuite(file) {
  const fullPath = path.join(__dirname, file);
  const mod = require(fullPath);

  // Look for the exported runner function:
  // `runUnitTests`, `runIntegrationTests`, `runPerformanceTests`, `runEdgeCasesTests`
  const runner = mod.runUnitTests
               || mod.runIntegrationTests
               || mod.runPerformanceTests
               || mod.runEdgeCasesTests;

  if (!runner) {
    throw new Error(`No test runner found in ${file}`);
  }

  console.log(`\n🧪 Running ${file}…`);
  await runner();
  console.log(`✅ ${file} passed`);
}

async function main() {
  const files = fs.readdirSync(__dirname)
    .filter(f => f.endsWith('.js') && f !== path.basename(__filename));

  let failed = 0;

  for (const file of files) {
    try {
      await runSuite(file);
    } catch (err) {
      console.error(`❌ ${file} failed:\n`, err.stack || err);
      failed++;
    }
  }

  console.log('\n📊 Summary:');
  if (failed) {
    console.error(`   🔴 ${failed} suite(s) failed`);
    process.exit(1);
  } else {
    console.log('   🎉 All suites passed!');
    process.exit(0);
  }
}

if (require.main === module) {
  main().catch(err => {
    console.error('Unexpected error:', err);
    process.exit(1);
  });
}
```

pkarr/pkg/test/edge-cases.js
```diff
@@ -0,0 +1,756 @@
...
+        const packet = builder.buildAndSign(keypair);
+        if (packet.records.length !== 3) {
+            throw new Error(`Expected 3 valid records, got ${packet.records.length}`);
+        }
+    });
+    
+    // Test 4: Very long record values
+    test("Very long record values", () => {
+        const builder = SignedPacket.builder();
+        const keypair = new Keypair();
```
SHAcollision: We could consider something like `pkg/test/helpers.js` to save on a lot of test code repetition.

```rust

const { Keypair, SignedPacketBuilder, Utils } = require('../pkarr.js');

/**
 * Spin up a fresh Keypair + Builder
 */
function newFixture() {
  const keypair = new Keypair();
  const builder = SignedPacketBuilder.builder();
  return { keypair, builder };
}

/**
 * Simple sleep/delay
 * @param {number} ms
 */
function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

module.exports = {
  newFixture,
  sleep,
  validatePublicKey: Utils.validatePublicKey,
};
```
use like
```
const { newFixture, sleep, validatePublicKey } = require('./helpers');
```

pkarr/pkg/README.md
```diff
@@ -0,0 +1,233 @@
...
+npm run test                  # Run all test suites (recommended)
+npm run test:unit            # Unit tests (29 tests) - Core functionality
+npm run test:integration     # Integration tests (7 tests) - Live network
+npm run test:performance     # Performance benchmarks - Speed analysis  
+npm run test:edge-cases      # Edge cases (21 tests) - Error handling
+```
+
+### Test Coverage
+
+- ✅ **Unit Tests** (29/29): Core WASM functionality, keypairs, packet building, API compatibility
```
SHAcollision: `(29/29)` ? Remove values from Readme.

pkarr/pkg/README.md
```diff
@@ -0,0 +1,233 @@
...
+- **`examples/advanced.js`** - Complex scenarios and best practices
+
+## 🔬 Testing
+
+Comprehensive test suite with 100% pass rate across all areas:
+
+```bash
+npm run test                  # Run all test suites (recommended)
+npm run test:unit            # Unit tests (29 tests) - Core functionality
+npm run test:integration     # Integration tests (7 tests) - Live network
```
SHAcollision: NIT: remove values `(7 tests) ` `(21 tests)`

pkarr/src/lib.rs
```diff
@@ -16,13 +16,15 @@ pub mod extra;
 mod keys;
 #[cfg(feature = "signed_packet")]
 mod signed_packet;
+#[cfg(all(target_family = "wasm", feature = "relays"))]
```
SHAcollision: If target is `wasm` feature should always be `relays` and only `relays` . Not sure if there is any way to enforce or disambiguate this.

pkarr/src/wasm/client.rs
```diff
@@ -0,0 +1,283 @@
...
+
+            relays
+                .publish(&signed_packet.inner, cas)
+                .await
+                .map_err(|e| ClientError::NetworkError(format!("publish failed: {}", e)))?;
+
+            Ok(())
+        }
+
+        #[cfg(not(feature = "relays"))]
```
SHAcollision: Ideally we should not need to deal with this if target is `wasm` .

pkarr/src/wasm/error.rs
```diff
@@ -0,0 +1,57 @@
+use wasm_bindgen::JsValue;
+
+/// Generic error types for the WASM module
+#[derive(Debug)]
+pub enum ClientError {
```
SHAcollision: How about using `thiserror` like
```rust
/// error.rs
use thiserror::Error;
use wasm_bindgen::prelude::*;

#[derive(Error, Debug)]
pub enum ClientError {
    #[error("validation failed in {context}: {message}")]
    Validation { context: String, message: String },

    #[error("parse error for {input}: {message}")]
    Parse { input: String, message: String },

    #[error("network error: {0}")]
    Network(String),

    #[error("configuration error: {0}")]
    Config(String),

    #[error("feature not enabled: {0}")]
    Feature(String),

    #[error("build error: {0}")]
    Build(String),
}

impl From<ClientError> for JsValue {
    fn from(e: ClientError) -> Self {
        JsValue::from_str(&e.to_string())
    }
}
```
so any time we add a variant it “just works.”

pkarr/src/wasm/client.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    /// Resolve a public key to get the latest signed packet
+    ///
+    /// # Arguments
+    /// * `public_key_str` - The public key as a z-base32 string
+    ///
+    /// # Returns
+    /// * `Option<SignedPacket>` - The signed packet if found
+    #[wasm_bindgen(js_name = "resolve")]
+    pub async fn resolve(
```
SHAcollision: NIT: just an idea to cut down on code lines.  AI generated macro.
```rust
    // Macro‐driven passthrough
    macro_rules! wrap {
        // wrap!(js_name, rust_name(arg1:Type1, arg2:Type2) -> RetTy);
        ($js:ident, $rust:ident($($arg:ident : $ty:ty),*) -> $ret:ty) => {
            #[wasm_bindgen(js_name = $js)]
            pub fn $rust(&self, $($arg: $ty),*) -> Result<$ret, JsValue> {
                self.inner.$rust($($arg),*).map_err(ClientError::from)
            }
        };
    }

    wrap!(publish, publish(&JsSignedPacket, Option<f64>) -> ());
    wrap!(resolve, resolve(&str) -> Option<JsSignedPacket>);
    wrap!(resolveMostRecent, resolve_most_recent(&str) -> Option<JsSignedPacket>);
```

pkarr/src/wasm/signed_packet.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    /// Check if the packet contains any records
+    #[wasm_bindgen(js_name = isEmpty)]
+    pub fn is_empty(&self) -> bool {
+        self.record_count() == 0
+    }
+
+    /// Get the DNS records as a JavaScript array of objects
+    #[wasm_bindgen(getter)]
+    pub fn records(&self) -> Result<js_sys::Array, JsValue> {
```
SHAcollision: Not sure if the record‐to‐JS‐object function could maybe be factored into a smaller helper something like:
```
fn set_prop(obj: &Object, key: &str, val: &JsValue) {
    Reflect::set(obj, &key.into(), val).unwrap();
}

fn record_to_js(rr: &ResourceRecord) -> Object {
    let o = Object::new();
    set_prop(&o, "name", &rr.name.to_string().into());
    // …etc…
    o
}

#[wasm_bindgen]
impl SignedPacket {
    #[wasm_bindgen(getter)]
    pub fn records(&self) -> Array {
        self.inner
            .all_resource_records()
            .map(|rr| record_to_js(&rr).into())
            .collect()
    }
}
```

pkarr/src/wasm/constants.rs
```diff
@@ -0,0 +1,77 @@
+//! Constants used throughout the WASM module
+
+// === Size and Time Constants ===
+
+/// Pkarr specification: packets must be <= 1000 bytes
+pub const MAX_PACKET_SIZE: usize = 1000;
```
SHAcollision: Why are we defining these `pub const` only under `/wasm` and not for borth `wasm` and `native`? For example, I see a lot of hardcoded `1000` in `src/signed_packet.rs`

SHAcollision (commented):
Hey @tipogi nice work here pushing the WASM pkarr forward.

Overall this is a solid end-to-end WASM packaging effort, but it feels like we’re hauling in thousands of lines of mostly AI generated boilerplate, docs, and generated tests. Some high-level thoughts:

- Core Rust bindings shouldn’t balloon with repetitive glue-code, maybe there is a way to use small helper functions or macros to remove the 1-for-1 `wasm_bindgen` wrappers and record serializers.

- Tests & examples are usually verbose, but almost every suite follows the same “builder+keypair+sleep+assert” pattern. Moving those into a helper (or a small table-driven loop) would help remove maybe even hundreds of lines.

Tests and docs can be verbose, but the core Rust+WASM layer should stay DRY macros or shared modules will help with maintainability.

pkarr/src/wasm/client.rs
tipogi: I don't know where you were suggesting to implement that 3 blocks of `wrap!` functions but macros inside an `impl` is rejected

Nuhvi:
This PR is getting way out of hand, so I think you should create a new directory (or an entirely different repo) for pkarr bindings, and add all of this in it. `pkarr` crate can be used as a dependency there, but I don't think any of this binding code should ever be shipped in the `pkarr` crate. I know I set the precedence in `pubky` crate, but things were more tightly integrated there, and the number of structs that needed bindings wasn't that hard.

One of the main reasons I never bothered with js bindings is that I didn't want to replicate every DNS type, so I am happy someone did all of that work, but let's move that work away from the `pkarr` rust crate, because I don't really care about the quality of JS libraries, so I will tend to do a bad job reviewing them, and sloppy reviews are safer when they are not touching files I am very protective of.

tipogi:
I hadn't thought about that, but it makes sense for me also. Let's separate the concerns and we can create another top folder with `bindings` name and under that, the crate folder which is the language-specific folder. In our case `js`. Do you think the folder naming is correct or do you have something different in mind?

Nuhvi:
Yeah `/bindings/js/` makes sense.

Nuhvi:
@tipogi Sorry for the intrusion, but I think you were not going in the right direction. I reverted all changes to the pkarr crate, and changed your code to use the top level pkarr client, because pkarr crate already ensures that you can use it in wasm without changing anything.

And the wasm workflow is meant to test pkarr crate in wasm alone, now we can change it diligently (no emojis and frivolous comments please!) to test the JS bindings, or we can create a separate workflow for the bindings, which is what I would prefer.

I haven't reviewed the bindings code by the way, so it might also contain room for improvement, or I might disagree with the API, but for now, I just wanted to get you out of your own way.

~Remains a cargo check issue, that I will deal with.~

tipogi:
@Nuhvi thanks for the help! when you are done, I can create another workflow to test the js binding as you describe

Nuhvi:
I am done it is passing as you can see.

bindings/js/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## 🔨 Building from Source
+
+To build the WASM package from source, run:
+
+```bash
+./build-wasm.sh
+
+# Test your changes
+cd pkg && npm run test
```
SHAcollision: The user would also need to to build and run the `pkarr-relay` in testnet mode, unless we improve the `npm run test` script to also always run the relay.

.github/workflows/js-binding.yaml
```diff
@@ -0,0 +1,65 @@
...
+        echo "Waiting for server to start..."
+        sleep 8
+        
+        # Change the working directory to JS bindings crate
+        cd ./bindings/js
+        
+        # Change to the pkg directory and run JavaScript tests
+        echo "Running JavaScript tests..."
+        cd pkg
+        npm run test
```
SHAcollision: Not sure if feaseable, but all of this could be turned into a `package.json` script `"npm run start-relay"` that runs `cargo run -p pkarr-relay -- --testnet` maybe? Then `npm run test` could become something like `npm run start-relay & node test/all.js` (some sleep could be required?).

Something like this maybe (untested):
```json
"scripts": {
  "start-relay": "cargo run -p pkarr-relay -- --testnet",
  "test": "sh -c \"npm run start-relay & pid=$!; sleep 8; node test/all.js; kill $pid\""
}
```
If this works, then the test instructions in Readme and the `js-binding.yaml` workflow can be simplified because `npm run test` will suffice.

bindings/js/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+
+# Required for WASM target
+[target.wasm32-unknown-unknown.dependencies]
+getrandom = { version = "0.2.15", default-features = false, features = ["js"] }
+
+[features]
+# JS bindings always require relays - make it a required feature
+default = ["relays"]
+relays = []
+
```
SHAcollision: Redundant feature declaration. `relays` is always needed for `wasm`,  I guess we could just remove this.
```suggestion
```

bindings/js/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Development Notes
+
+The JavaScript bindings require the `relays` feature to function, as WASM environments cannot access the DHT directly. If you need to build with `--no-default-features`, you must explicitly enable the relays feature:
+
+```bash
+cargo check --no-default-features --features relays
+cargo build --no-default-features --features relays
+```
+
```
SHAcollision: We don't need `relays` to be a feature of crate `bindings/js` , if so, we could remove all of this as well.

bindings/js/README.md
```diff
@@ -0,0 +1,158 @@
...
+- **`Client`** - Publish and resolve packets via Pubky relays
+- **`Keypair`** - Generate and manage Ed25519 keypairs
+- **`SignedPacket`** - Build and sign DNS packets
+- **`Utils`** - Utility functions for validation
+
+### Client Methods
+
+```javascript
+const client = new Client();                    // Default relays
```
SHAcollision, line 66: Example missing the `import` statement (what is `Client()` and where does it come from).

bindings/js/README.md
```diff
@@ -0,0 +1,158 @@
...
+builder.addCnameRecord(name, target, ttl);      // Canonical names
+builder.addHttpsRecord(name, priority, target, ttl);  // HTTPS service records
+builder.addSvcbRecord(name, priority, target, ttl);   // Service binding records
+builder.addNsRecord(name, nameserver, ttl);     // Name server records
+
+// Optional: Set custom timestamp
+builder.setTimestamp(Date.now());
+
+// Build and sign
+const packet = builder.buildAndSign(keypair);
```
SHAcollision: nice!

bindings/js/src/signed_packet.rs
```diff
@@ -0,0 +1,258 @@
...
+/// WASM-compatible wrapper for SignedPacket
+#[wasm_bindgen]
+pub struct SignedPacket {
+    pub(super) inner: pkarr::SignedPacket,
+}
+
+#[wasm_bindgen]
+impl SignedPacket {
+    /// Get the public key as a z-base32 string
+    #[wasm_bindgen(getter, js_name = publicKeyString)]
```
SHAcollision, line 14: nice getters

bindings/js/src/signed_packet.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// Get the DNS records as a JavaScript array of objects
+    #[wasm_bindgen(getter)]
+    pub fn records(&self) -> Result<js_sys::Array, JsValue> {
+        let records = js_sys::Array::new();
+
+        for record in self.inner.all_resource_records() {
+            if let Ok(js_record) = Self::record_to_js(record) {
+                records.push(&js_record);
+            }
+            // Skip unsupported record types silently
```
SHAcollision, line 46: Nit: maybe it should log something?

bindings/js/src/constants.rs
```diff
@@ -0,0 +1,77 @@
+//! Constants used throughout the WASM module
```
SHAcollision, line 1: I still think many of these constant should be shared (or directly defined) on the `native` pkarr, having them defined twice could lead to drift in behaviour native vs wasm. Maybe work for a future PR if this PR is meant to change nothing on the `pkarr` crate.

SHAcollision (commented):
At a high level it looks good: it works and it's highly functional, it is well tested, structure makes sense and it's readable and well documented. Passing the ball to @Nuhvi for a more in depth check.

.github/workflows/js-binding.yaml
tipogi: it seems `sh -c` has some problems to catch the testnet relay _PID_, going with `node -e` able to catch and kill the process

bindings/js/README.md
tipogi: https://github.com/pubky/pkarr/pull/153/commits/ae8d4f61c8e76b614fe27b8b9b3501bd25c2f4df

.github/workflows/js-binding.yaml
tipogi: https://github.com/pubky/pkarr/pull/153/commits/ae8d4f61c8e76b614fe27b8b9b3501bd25c2f4df

bindings/js/README.md
tipogi: https://github.com/pubky/pkarr/pull/153/commits/ae8d4f61c8e76b614fe27b8b9b3501bd25c2f4df

bindings/js/README.md
tipogi: https://github.com/pubky/pkarr/pull/153/commits/ae8d4f61c8e76b614fe27b8b9b3501bd25c2f4df

bindings/js/Cargo.toml
tipogi: https://github.com/pubky/pkarr/pull/153/commits/e774d0192698cfe48e47d569d6c694a23b98eec2

tipogi:
@Nuhvi do you approve that PR? or are you short in time?

Nuhvi:
I am short on time, but I also left notes about what I think is redundant modules. If you want me to cleanup that I can do it when I have some time.

tipogi:
> I don't really care about the quality of JS libraries

Referring to this text, I think then we can merge. Thanks for your time

PR #165, opened 2025-07-22
chore: binding build target changed

tipogi:
It was an issue that the WASM build was targeting **Node.js** specifically (`--target nodejs`), which generated code that used `Node.js`-specific modules like `fs` and `path`. This caused the error in the browser.

#### ❌ Attempt
Updated build target from `--target nodejs` to `--target bundler`. Now the binding is compatible with modern bundlers like webpack (used by Next.js) but in nodeJS was throwing errors..

#### ✅ Solution
Cannot use fully `wasm-pack` to build the binding because it seems there is some problems creating a compatible bundle for _web_ and _nodejs_. Finally used that solution that it is pointing in [this issue](https://github.com/rustwasm/wasm-pack/issues/1334)


bindings/js/pkg/example.js
```diff
@@ -134,11 +137,13 @@ async function runExample() {
         console.log(`   ✅ Default relays: ${defaultRelays.length} relays`);
         
         // Packet serialization and parsing
-        const packetBytes = signedPacket.toBytes();
-        console.log(`   📦 Packet size: ${packetBytes.length} bytes`);
+        const packetBytes = signedPacket.uncompressedBytes();
```
SeverinAlexB: Maybe `signedPacket.bytes()` is better than `uncompressedBytes()`? I know I proposed the uncompressed but looking at it, `bytes()` sounds better.

SeverinAlexB (approved):
One NIT, otherwise good to go!

*I didn't run the code

bindings/js/pkg/example.js
tipogi: https://github.com/pubky/pkarr/pull/165/commits/3fdea313257bd0fa2612491dcdf08601c0e8099f

PR #166, opened 2025-07-23
`Keypair`: add `from_secret_key_file`, `write_secret_key_file`

ok300:
This PR introduces an optional filesystem utility for `pkarr` clients, which allows them to write a keypair to disk.

It essentially brings this operation from the homeserver

https://github.com/pubky/pubky-core/blob/880c696dd2e0b1a4c1df08602ea6cec1a35cbb6f/pubky-homeserver/src/data_directory/persistent_data_dir.rs#L68-L79

to `pkarr`.

The reasons for this are:

- the peculiar `cfg(unix)` permissions handling, which someone integrating `pkarr` might not consider when persisting keypairs to disk
- code re-use: in addition to the homeserver, this same operation is now needed on the Nexus indexer side, so placing this utility method here avoids code duplication
- intuitive location: since it specifically refers to handling `pkarr:Keypair`s, the `extra` optional module in `pkarr` seems like a good place for it


pkarr/src/extra/fs.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+use std::path::PathBuf;
+
+use crate::Keypair;
+
+/// Writes the secret key of the keypair to file, as a hex encoded string.
+/// If the file already exists, it will be overwritten.
+pub fn write_keypair(keypair: &Keypair, secret_file_path: PathBuf) -> Result<(), std::io::Error> {
```
SeverinAlexB: Use `&Path` instead of `PathBuf`.

pkarr/src/extra/fs.rs
```diff
@@ -0,0 +1,21 @@
...
+    let secret = keypair.secret_key();
+    let hex_string = const_hex::encode(secret);
+    std::fs::write(secret_file_path.clone(), hex_string)?;
+    #[cfg(unix)]
+    {
+        std::fs::set_permissions(&secret_file_path, std::fs::Permissions::from_mode(0o600))?;
+    }
+    tracing::info!("Secret file created at {}", secret_file_path.display());
```
SeverinAlexB: No logs in utility functions

pkarr/Cargo.toml
```diff
@@ -46,6 +46,9 @@ tracing = { version = "0.1.41", optional = true }
 # feat: dht dependencies
 mainline = { version = "5.4.0", optional = true }
 
+# feat: fs dependencies
+const-hex = { version = "1.14", optional = true }
```
SeverinAlexB: Using `const-hex` will bloat the library with an additional 22KB compared to just using the std lib. Removing `const-hex` will also simplify the lib as the `fs` feature flag is not needed anymore.

https://grok.com/share/bGVnYWN5_cb038f99-326f-414e-a88b-8a1b6d19aca9

pkarr/src/extra/fs.rs
```diff
@@ -0,0 +1,21 @@
+//! Filesystem operations involving pkarr keys.
```
SeverinAlexB: If we don't gate these features behind `fs` anymore, we can move these functions either to `src/keys.rs` or `src/utils.rs`. Up to you.

SeverinAlexB (changes requested):
I like the idea of adding these utility function here. If we add `write_keypair()`, we should also add a `read_keypair()` function.

pkarr/src/extra/fs.rs
ok300: Done: fddc10d91f7a7d3fa514c1f6d3e7dd447d4173c1

pkarr/src/extra/fs.rs
ok300: Done: 43ce79880855742359798263380fb7d3a317ae2b

pkarr/Cargo.toml
ok300: Good idea, done: e21390d07cd9765bda3cbc5b121254fa32cf24c7

pkarr/src/extra/fs.rs
ok300: Done: 6f8fe6148a0c59fd88906fe9fe70f6a571042bce

ok300:
> If we add write_keypair(), we should also add a read_keypair() function.

Done.

Combined with the moving to `keys.rs`, a slight renaming seemed appropriate: 
- `write_keypair` -> `Keypair::write_secret_key_file`
- `read_keypair` -> `Keypair::from_secret_key_file`

ok300:
Ready for review.

The failing CI seems unrelated to this change. Instead, @SHAcollision suggests it might be caused by recent changes like #165 .

Even if this gets ACKed, I will hold back the merge until the CI pipeline is fixed.

SHAcollision:
> The failing CI seems unrelated to this change.

Seems to have been some unrelated temporal issue. Some checks in the `pubky-core` repo were also failing with same error. They are all now passing.

pkarr/src/keys.rs
```diff
@@ -63,6 +63,53 @@ impl Keypair {
...
+            let byte = u8::from_str_radix(byte_str, 16)?;
+            secret_key_bytes_vec.push(byte);
+        }
+
+        let secret_key_bytes: [u8; 32] = secret_key_bytes_vec
+            .try_into()
+            .map_err(|_| "Invalid secret key length")?;
+
+        Ok(Keypair::from_secret_key(&secret_key_bytes))
+    }
```
SeverinAlexB: If you use `std::io::ErrorKind::InvalidData` we can return `std::io::Error` instead of the weird `Box<dyn std::error::Error>`.

`InvalidData: Data not valid for the operation were encountered.` does fit well with our purpose.

```rust
    /// Reads the `SecretKey` from a file and derives the `Keypair` from it.
    pub fn from_secret_key_file(
        secret_file_path: &std::path::Path,
    ) -> Result<Keypair, std::io::Error> {
        let hex_string = std::fs::read_to_string(secret_file_path)?;
        let hex_string = hex_string.trim();

        if hex_string.len() % 2 != 0 {
            return Err(std::io::Error::new(std::io::ErrorKind::InvalidData, "Invalid hex string length"));
        }

        let mut secret_key_bytes_vec = vec![];
        for i in (0..hex_string.len()).step_by(2) {
            let byte_str = &hex_string[i..i + 2];
            let byte = u8::from_str_radix(byte_str, 16).map_err(|_| std::io::Error::new(std::io::ErrorKind::InvalidData, "Invalid hex string"))?;
            secret_key_bytes_vec.push(byte);
        }

        let secret_key_bytes: [u8; 32] = secret_key_bytes_vec
            .try_into()
            .map_err(|_| std::io::Error::new(std::io::ErrorKind::InvalidData, "Invalid secret key length"))?;

        Ok(Keypair::from_secret_key(&secret_key_bytes))
    }
```

pkarr/src/keys.rs
```diff
@@ -63,6 +63,53 @@ impl Keypair {
     }
 }
 
+// Filesystem-related operations, which are not available for WASM
+#[cfg(not(wasm_browser))]
+impl Keypair {
+    /// Reads the `SecretKey` from a file and derives the `Keypair` from it.
```
SeverinAlexB: NIT

```suggestion
    /// Reads the `SecretKey` from a hex file and derives the `Keypair` from it.
```

pkarr/src/keys.rs
```diff
@@ -63,6 +63,53 @@ impl Keypair {
...
+
+        let secret_key_bytes: [u8; 32] = secret_key_bytes_vec
+            .try_into()
+            .map_err(|_| "Invalid secret key length")?;
+
+        Ok(Keypair::from_secret_key(&secret_key_bytes))
+    }
+
+    /// Writes the secret key of the keypair to file, as a hex encoded string.
+    /// If the file already exists, it will be overwritten.
```
SeverinAlexB:
```suggestion
    /// Writes the secret of the keypair to a file, as a hex encoded string.
    /// If the file already exists, it will be overwritten.
    /// In unix like operating systems, the file permission `600` is set.
```

pkarr/src/keys.rs
```diff
@@ -557,4 +604,59 @@ mod tests {
...
+            assert!(write_keypair_result.is_ok());
+            assert!(temp_file_path.exists());
+
+            let read_keypair_result = Keypair::from_secret_key_file(&temp_file_path);
+            assert!(read_keypair_result.is_ok());
+
+            let read_keypair = read_keypair_result.unwrap();
+            assert_eq!(generated_keypair.secret_key(), read_keypair.secret_key());
+
+            let _ = remove_file(&temp_file_path);
```
SeverinAlexB: NIT: This test will leave temp files when assertions fail. Use https://crates.io/crates/tempfile as a dev dependency

SeverinAlexB (changes requested):
Almost good

pkarr/src/keys.rs
ok300: Good idea, done: c3764c291376e39762c3195f15c02eafb2e1f332

pkarr/src/keys.rs
ok300: Done: 65b5797d3b23fd005db8efdc1b009611d10e77eb

SeverinAlexB (approved):
LGTM

SHAcollision (approved):

Issue #177, opened 2025-08-04
Update postcard to v0.2 when available/stable


SeverinAlexB:
postcard uses `atomic-polyfill v1.0.3` indirectly via `heapless v0.7.17`. atomic-polyfill is unmaintained and should be upgraded. 

We should update as soon as postcard v0.2 is stable.

https://github.com/jamesmunns/postcard/issues/223
https://github.com/jamesmunns/postcard/issues/241

PR #178, opened 2025-08-04
chore: cargo deny

SeverinAlexB:
Adds [cargo deny](https://github.com/EmbarkStudios/cargo-deny) to the CI.
- Checks if any dependency has a unwanted license.
- Check for security advisories from https://rustsec.org/advisories/ like https://github.com/pubky/pkarr/issues/177


SHAcollision (approved):
Looks good

PR #179, opened 2025-08-06
fix(pkarr): &String.try_into::<PublicKey>() stack overflow

SeverinAlexB:
Fixes a stack overflow in `impl TryFrom<&String> for PublicKey`. Previously, it would call itself over and over again which leads to a panic.


SHAcollision (approved):
Nice catch

PR #183, opened 2025-08-21
Contribution doc, ai-generated first draft

BitcoinErrorLog:
This is an AI-generated first draft, assume it is bad, but please use it to instigate and publish something not bad.


SeverinAlexB:
I would like to close this. I don't think we should gear other people even more towards contribution especially in this AI day and age. It's likely more of a waste of time than it's worth.

PR #184, opened 2025-09-04
Add new relay modes

afterburn:
This PR adds 3 new modes for the relay:
1. Legacy - Same behavior as before, uses cache and rate limiting, ignores new most_recent query parameter
2. Public - Uses cache and rate limiting, respects new `most_recent` query parameter which will skip cache.
3. Private - Same as public, but allows rate limiting bypass if IP is whitelisted

### DIFF
--- a/relay/src/config.example.toml
+++ b/relay/src/config.example.toml
@@ -1,3 +1,9 @@
+# Relay operating mode
+# - legacy: ignores fresh query parameter, always returns cached data
+# - public: ignores fresh query parameter, always returns cached data (default)
+# - private: when fresh=true specified, queries DHT for most recent data
+mode = "public"
+
 # HTTP server configurations.
 [http]
 # The port number to run the HTTP server on. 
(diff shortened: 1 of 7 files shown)


relay/src/config.example.toml
```diff
@@ -1,3 +1,9 @@
+# Relay operating mode
+# - legacy: ignores fresh query parameter, always returns cached data
+# - public: ignores fresh query parameter, always returns cached data (default)
+# - private: when fresh=true specified, queries DHT for most recent data
```
SeverinAlexB, line 4: I don't think legacy is needed. `public` and `private is sufficient because they should be backwards compatible.

Another note: Is the naming `public/private` good? The difference between them is the support for `resolve_most_recent`. I am not sure if Public/Private fits. What do you think?

relay/src/config.example.toml
```diff
@@ -41,3 +47,8 @@ behind_proxy = false
 burst_size = 10
 # Number of seconds after which one request of the quota is replenished.
 per_second = 2
+
+# Optional IP addresses that bypass rate limiting when running in private mode.
+# Example:
+# ip_whitelist = ["192.168.1.100", "10.0.0.0/8", "172.16.0.0/12"]
+ip_whitelist = []
```
SeverinAlexB, line 54: Can this be a whitelist for the `public` mode too?

SeverinAlexB (commented):
So far so good.

PR #185, opened 2025-09-18
feat(relay): Operation-based rate limiting + Resolve most recent

afterburn:
**Summary**
This PR replaces current rate limiting with an operation-based configuration that maps directly to the relay's three core operations: `publish`, `resolve`, and `resolve_most_recent`. 

**Configuration Example**
```toml
[relay]
[[relay.rate_limits]]
operation = "resolve"
quota = "10r/s"
burst = 20

[[relay.rate_limits]]
operation = "resolve_most_recent"
quota = "2r/s"
burst = 5

[[relay.rate_limits]]
operation = "publish"
quota = "2r/s"
burst = 10
whitelist = ["127.0.0.1", "192.168.1.0/24"]  # optional, both ips and subnets supported
```

**Operations in this toml file automatically map to:**
- publish → PUT requests (publishing records)
- resolve → GET requests without most_recent=true (cached lookups)
- resolve_most_recent → GET requests with most_recent=true (DHT queries)

If no config.toml file is present, system will use defaults, so this is not a breaking change.


relay/config.example.toml
```diff
@@ -0,0 +1,58 @@
...
+maximum_ttl = 86400
+
+# Operation-based rate limiting configurations.
+# If not included, rate limiting will use defaults.
+[relay]
+# Rate limits are applied to specific relay operations
+# Each rate limit entry supports:
+# - operation: one of "publish", "resolve", "resolve_most_recent"
+# - quota: rate in format "{count}{unit}/{time}" (e.g., "2r/s", "100mb/m")
+#   - units: "r" (requests), "kb", "mb", "gb"
```
SeverinAlexB: kb, mb, gb don't make sense in this relay usecase

relay/config.example.toml
```diff
@@ -0,0 +1,58 @@
...
+quota = "2r/s"
+burst = 5
+
+# Rate limit publish operations (PUT requests) - very restrictive with whitelist
+[[relay.rate_limits]]
+operation = "publish"
+quota = "1r/s"
+burst = 2
+# Per-operation whitelist - supports individual IPs and CIDR blocks
+whitelist = ["127.0.0.1", "192.168.1.0/24", "10.0.0.0/8"]
```
SeverinAlexB: Most users will apply the examples values. Let's use sane default values for public relays here.

This is not a perfect science so let me know if you think these are not good. Suggestions:

- resolve -> 30r/m burst 30
- resolve_most_recent -> 5r/m burst 10
- publish -> 5r/m burst 10

relay/src/rate_limiter/extract_ip.rs
```diff
@@ -0,0 +1,37 @@
...
+        .map(|addr| addr.ip())
+}
+
+pub fn extract_ip<T>(req: &Request<T>) -> anyhow::Result<IpAddr> {
+    let headers = req.headers();
+    maybe_x_forwarded_for(headers)
+        .or_else(|| maybe_x_real_ip(headers))
+        .or_else(|| maybe_connect_info(req))
+        .ok_or(anyhow::anyhow!("Failed to extract ip."))
+}
```
SeverinAlexB: Just realized there is an attack vector in this. Users can send spoofed `x-forwarded-for` or `x-real-ip` headers and get around the rate limiting. Please have a look at this here and fix it in the homeserver too.

`x-forwarded-for` and `x-real-ip` are used when the relay is behind a load balancer like nginx.

relay/src/rate_limiter/layer.rs
```diff
@@ -0,0 +1,748 @@
...
+                            Some("Failed to extract key for rate limiting"),
+                        )
+                        .into_response())
+                    });
+                }
+            };
+
+            // Check whitelist using the original IP address
+            let ip = extract_ip(&req)
+                .unwrap_or_else(|_| "127.0.0.1".parse().expect("localhost IP is valid"));
```
SeverinAlexB: What's the purpose of extracting the requests IP address twice here? Once with limit.extract_key and once with extract_ip

relay/src/rate_limiter/layer.rs
```diff
@@ -0,0 +1,748 @@
...
+        // Match the request to the defined limits.
+        let limits = self.get_limit_matches(&req);
+        if limits.is_empty() {
+            // No limits matched, so we can just call the next layer.
+            return Box::pin(async move { inner.call(req).await.map_err(|_| unreachable!()) });
+        }
+
+        // Go through all the limits and check if we need to throttle or reject the request.
+        for limit in limits.clone() {
+            let key = match limit.extract_key(&req) {
```
SeverinAlexB: Just calling `GET http://0.0.0.0:6881/` matches the resolve rate limiter. This shouldn't be the case.

SeverinAlexB (changes requested):
Need to do a second round later

afterburn:
@SeverinAlexB resolved your feedback:
- Removed kb, mb, gb quota units, now always uses either r/s or r/m as requested
- Changed default rate limits to something saner
- Added `behind_proxy` flag in config toml, which allows users to respect/ignore proxy headers
- Cleaned up IP extraction, only happens once per request now
- Added operation 'Index' so users can specify rate limiting for root path

relay/src/config.rs
```diff
@@ -59,18 +69,54 @@ pub struct Config {
...
+            },
+        ];
+
         let mut this = Self {
             http_port: 6881,
             pkarr: Default::default(),
             cache_path: None,
             cache_size: DEFAULT_CACHE_SIZE,
-            rate_limiter: Some(RateLimiterConfig::default()),
+            rate_limiter: Some(default_rate_limits),
```
SeverinAlexB: If no rate limits are defined in the config, no rate limits should be applied. This is not the case though, is it?

relay/src/handlers.rs
```diff
@@ -63,12 +66,19 @@ pub async fn put(
...
 ) -> Result<impl IntoResponse, Error> {
     let public_key = PublicKey::try_from(public_key.as_str())
         .map_err(|error| Error::new(StatusCode::BAD_REQUEST, Some(error)))?;
 
-    if let Some(signed_packet) = state.client.resolve(&public_key).await {
+    let signed_packet = if query.most_recent {
+        state.client.resolve_most_recent(&public_key).await
+    } else {
+        state.client.resolve(&public_key).await
+    };
```
SeverinAlexB: Do we need a way to fully disable resolve_most_recent? A public relay might not want to allow it at all?

What do you think? Do we need a `resolve_most_recent_enabled = true/false` config?

SeverinAlexB (commented):
Two comments, otherwise looks good.

In the Pubky Core team, we have a convention on how to format PR title. We stick to the [conventional commit format](https://www.conventionalcommits.org/en/v1.0.0/).

So for this PR, the title could be: feat(relay): Operation-based rate limiting + Resolve most recent


CleanShot 2025-09-25 at 10 37 51@2x

afterburn:
@SeverinAlexB 

- If no rate limits are defined in `config.toml`, rate limiting will be disabled completely
- Added `Index` to `Operation` enum so calling `GET http://0.0.0.0:6881/` doesn't trigger the incorrect rate limit
- Added `resolve_most_recent_enabled` flag so users have the ability to completely ignore the `most_recent=true` query parameter (defaults to `false` for backward compatibility)

relay/src/config.rs
```diff
@@ -4,22 +4,33 @@ use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 use std::{
     fmt::Debug,
+    num::NonZero,
     path::{Path, PathBuf},
+    str::FromStr,
```
SeverinAlexB: It seems like NonZero and FromStr and not used. Can you double check the Github workflow and why the format/clippy check doesn't fail?

SeverinAlexB (commented):
One NIT, otherwise good to go

SeverinAlexB:
Actually, should this add the `most_recent` option to the pkarr client too? You are just adding it to the relay in this PR?

pkarr/src/client/builder.rs
```diff
@@ -49,6 +49,14 @@ pub(crate) struct Config {
     /// Defaults to [DEFAULT_REQUEST_TIMEOUT]
     pub request_timeout: Duration,
 
+    /// Whether to pass the most_recent query parameter to relay requests
+    ///
+    /// If true, relays will be asked to bypass cache and return the most recent packet.
+    /// Note: Relays must also have resolve_most_recent_enabled=true to honor this.
+    ///
+    /// Defaults to false for backward compatibility.
+    pub resolve_most_recent_enabled: bool,
```
dzdidi: why not just `resolve_most_recent` ? and then method for `set_resolve_most_recent(value: boolean)` or even two methods `enable_resolve_most_recent` and `disable_resolve_most_recent` ?

pkarr/src/client/builder.rs
```diff
@@ -268,6 +277,18 @@ impl ClientBuilder {
         self
     }
 
+    /// Set whether to pass the most_recent query parameter to relay requests.
+    ///
+    /// If enabled, relays will be asked to bypass cache and return the most recent packet.
+    /// Note: Relays must also have resolve_most_recent_enabled=true to honor this.
+    ///
+    /// Defaults to false.
+    pub fn resolve_most_recent_enabled(&mut self, enabled: bool) -> &mut Self {
```
dzdidi: This is name of a variable or of a getter but not a setter function. Prefer to name methods that start with verbs which identify actions that method is performing

relay/src/config.rs
```diff
@@ -7,19 +7,28 @@ use std::{
...
 pub const CACHE_DIR: &str = "pkarr-cache";
 
-use crate::rate_limiting::RateLimiterConfig;
-
 #[derive(Serialize, Deserialize, Default)]
 struct ConfigToml {
     http: Option<HttpConfig>,
     mainline: Option<MainlineConfig>,
-    rate_limiter: Option<RateLimiterConfig>,
+    relay: Option<RelayToml>,
```
dzdidi: Why break consistency and not calling it `RelayConfig`?

relay/src/config.rs
```diff
@@ -127,7 +152,23 @@ impl Config {
             config.http_port = port;
         }
 
-        config.rate_limiter = config_toml.rate_limiter;
+        // Apply relay configuration
+        if let Some(relay_config) = config_toml.relay {
+            // Only apply rate limits if explicitly defined in TOML
+            if let Some(rate_limits) = relay_config.rate_limits {
+                config.rate_limiter = Some(rate_limits);
+            }
```
dzdidi:
```suggestion
            config.rate_limits = relay_config.rate_limits.unwrap_or(config.rate_limits);
```

dzdidi: or `map`

relay/Cargo.toml
```diff
@@ -30,16 +30,23 @@ thiserror = "2.0.11"
 bytes = "1.10.0"
 tower_governor = "0.6.0"
 governor = "0.8.0"
+tower = "0.5.0"
+futures-util = "0.3"
+regex = "1.0"
```
SHAcollision: This dep does not seem to be used.

relay/src/rate_limiter/layer.rs
```diff
@@ -0,0 +1,594 @@
...
+            Operation::Publish => {
+                req.method() == Method::PUT && Self::is_operation(req.uri().path())
+            }
+            Operation::Resolve => {
+                if req.method() != Method::GET || !Self::is_operation(req.uri().path()) {
+                    return false;
+                }
+                // Check if most_recent query parameter is present and true
+                if let Some(query_string) = req.uri().query() {
+                    !query_string.contains("most_recent=true")
```
SHAcollision: Nit: here and on L327,  `.contains()` is brittle and misclassifies inputs like `foo=almost_recent=true` or URL-encoded variants. 

We could parse the query properly, e.g. via 
```rust
url::form_urlencoded::parse(req.uri().query().unwrap_or_default().as_bytes())`
```
Or the Axum native Query, what ever is easier. Then test key/value equality.

We could build a tiny helper `fn is_most_recent(req: &Request<Body>) -> bool` and use here and on line 327, that would make this `match` more readable.

pkarr/src/client.rs
```diff
@@ -52,6 +52,7 @@ pub(crate) struct Inner {
     dht: Option<Dht>,
     #[cfg(relays)]
     relays: Option<RelaysClient>,
+    resolve_most_recent_enabled: bool,
```
SHAcollision: Is this ever used?

relay/src/handlers.rs
```diff
@@ -63,12 +66,19 @@ pub async fn put(
...
     State(state): State<AppState>,
     Path(public_key): Path<String>,
+    Query(query): Query<GetQuery>,
     request_headers: HeaderMap,
 ) -> Result<impl IntoResponse, Error> {
     let public_key = PublicKey::try_from(public_key.as_str())
         .map_err(|error| Error::new(StatusCode::BAD_REQUEST, Some(error)))?;
 
-    if let Some(signed_packet) = state.client.resolve(&public_key).await {
+    let signed_packet = if query.most_recent && state.resolve_most_recent_enabled {
```
SHAcollision: If I am not mistaken, the handler guards the bypass of cache with `state.resolve_most_recent_enabled`, but the rate-limiter will still treat `?most_recent=true` as “resolve_most_recent” (even if it is disabled) and apply the stricter quota even when the feature is disabled. Tie classification to the same flag to avoid surprising throttling when the feature is off.

relay/src/lib.rs
```diff
@@ -139,21 +145,10 @@ impl Relay {
...
-        let rate_limiter = config
-            .rate_limiter
-            .map(|rate_limiter| rate_limiting::IpRateLimiter::new(&rate_limiter));
+        let rate_limiter = config.rate_limiter.clone();
 
         config.pkarr.cache(cache);
 
-        if let Some(ref rate_limiter) = rate_limiter {
-            config.pkarr.dht(|builder| {
-                builder.server_settings(pkarr::mainline::ServerSettings {
```
SHAcollision, line 155: Regression maybe? The prior implementation also wired the rate-limiter into the DHT node via `ServerSettings{ filter: … }`, providing protection at the DHT layer. The new code drops that hook, so only HTTP is rate-limited now. Is his corrrect? If so, either preserve a DHT‐level filter or document the intentional removal.

SHAcollision (commented):
Great progress! Left a few minor comments and nits.

afterburn:
@SeverinAlexB @SHAcollision Restored DHT rate limiting that was previously removed. Since the DHT can be accessed directly, keeping rate limiting in place is necessary. The implementation is now fully backward compatible. Existing `[rate_limiter]` configurations continue to work as before. New `[dht_rate_limit]` and `[http_rate_limit]` options are opt-in. By default, it falls back to the legacy IP-based limiter unless one of the new configs is defined. Also added a comprehensive test to cover the new behavior and ensure backward compat.

relay/src/config.rs
```diff
@@ -59,8 +78,24 @@ pub struct Config {
...
+    ///
+    /// If true, trusts X-Forwarded-For and X-Real-IP headers.
+    /// If false, only uses the direct TCP connection IP.
+    ///
+    /// Defaults to false for security.
+    pub behind_proxy: bool,
+    /// Whether to respect the most_recent query parameter in resolve requests
+    ///
+    /// If true, the most_recent query parameter is respected and will bypass cache.
+    /// If false, the most_recent query parameter is ignored and will always load from cache.
```
86667, line 95: This feels like difficult behaviour to debug. Did we consider returning an error if `most_recent` is requested but the relay is configured with `resolve_most_recent: false`?

relay/src/lib.rs
```diff
@@ -68,16 +72,26 @@ impl RelayBuilder {
...
+        self.0.dht_rate_limiter = None;
+        self.0.http_rate_limiter = None;
+
+        self
+    }
+
+    /// Set the operation-based rate limiter configuration.
+    ///
+    /// Defaults to None (no operation-based rate limiting).
+    pub fn rate_limiter_config(&mut self, limits: Option<Vec<OperationLimit>>) -> &mut Self {
```
86667, line 84: This may technically be a breaking change. im not sure if we have any other consumers of the module as a lib

pkarr/Cargo.toml
```diff
@@ -1,12 +1,13 @@
 [package]
 name = "pkarr"
-version = "5.0.3"
+version = "6.0.0-rc.0"
```
86667, line 3: Should we be merging into `v6` branch

86667: if so, maybe we can remove the config backwards compatibility logic

relay/src/rate_limiter/layer.rs
```diff
@@ -0,0 +1,598 @@
...
+                        StatusCode::BAD_REQUEST,
+                        Some("Failed to create rate limiting key"),
+                    )
+                    .into_response())
+                });
+            }
+        };
+
+        // Go through all the limits and check if we need to throttle or reject the request.
+        for limit in limits.clone() {
```
86667, line 575: nit: no need to clone

relay/src/rate_limiter/layer.rs
```diff
@@ -0,0 +1,598 @@
...
+    /// Check if a request matches this operation
+    pub fn matches_request(&self, req: &Request<Body>, resolve_most_recent: bool) -> bool {
+        match self {
+            Operation::Publish => {
+                req.method() == Method::PUT && Self::is_operation(req.uri().path())
+            }
+            Operation::Resolve => {
+                req.method() == Method::GET
+                    && Self::is_operation(req.uri().path())
+                    && !(resolve_most_recent && Self::has_query_param(req, "most_recent", "true"))
```
86667, line 321: Why does the rate limiter need to consider the `resolve_most_recent` config var here? it cares for rate limiting based on the requests, the process's config is irrelevant

relay/src/rate_limiter/layer.rs
```diff
@@ -0,0 +1,598 @@
...
+            if limit.limit.is_whitelisted(&ip) {
+                continue;
+            }
+
+            // Check request rate limit
+            if let Err(_e) = limit.limiter.check_key(&key) {
+                tracing::debug!("Rate limit of {} exceeded for {:?}", limit.limit.quota, key);
+                return Box::pin(async move {
+                    Ok(
+                        HttpError::new(StatusCode::TOO_MANY_REQUESTS, Some("Rate limit exceeded"))
```
86667, line 586: `Retry-After` header always nice for 429s

pkarr/Cargo.toml
86667: especially concerning is `behind_proxy` being in two possible places - which takes precedence? If we need to keep backwards compatibility then this should be documented. Quick AI check says that `[relay].behind_proxy` will infact override `[rate_limiter].behind_proxy`

86667 (commented):
Few comments 👍 

Also we could have quite a few more unit tests. All behaviour may be tested somewhere but unit tests inline with the code they test are pretty much always worth it these days with how fast AI can write them IMO

PR #190, opened 2025-10-03
chore: v5.0.0 + mainline v6.0.0

SeverinAlexB:
Updates pkarr and the relay to mainline v6.0.0.

New major version required because mainline dep update is a major version and mainline is exported in pkarr.

FYI: @dignifiedquire


dzdidi (approved):

dignifiedquire:
thanks!

PR #195, opened 2025-11-10
feat: Improve multi arch builds for pkarr

SpontaneousOverthrow:
currently the build arg `TARGETARCH=aarch64` should be provided in order to build image for m1 macs.  
by default the build arg `TARGETARCH` exists, but for m1 macs it's equal `arm64`
I have tested this PR on m1 mac and x86_64 linux machine and it's working.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 FROM rust:1.86.0-alpine3.20 AS builder
 
 # Build platform argument (x86_64 or aarch64) (default: x86_64)
-ARG TARGETARCH=x86_64
+ARG TARGETARCH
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@
 FROM rust:1.86.0-alpine3.20 AS builder
 
 # Build platform argument (x86_64 or aarch64) (default: x86_64)
-ARG TARGETARCH=x86_64
+ARG TARGETARCH
```
catch-21, line 7: can remove all `TARGETARCH`

catch-21 (approved):

PR #198, opened 2026-01-08
fix: unmantained rustls-pemfile deny CI failure. patch version bump.

SHAcollision:
**WIP**
- Upgraded `axum-server` to 0.8 in the relay crate and pkarr dev-dependencies to drop the unmaintained `rustls-pemfile` chain.
- Updated for the new server behavior.
- Bumped `pkarr` and `pkarr-relay` patch versions for a new release.

### DIFF
--- a/pkarr/examples/http-get.rs
+++ b/pkarr/examples/http-get.rs
@@ -2,7 +2,6 @@
 
 use reqwest::Method;
 use tracing::Level;
-use tracing_subscriber;
 
 use clap::Parser;
 
--- a/pkarr/examples/http-serve.rs
+++ b/pkarr/examples/http-serve.rs
@@ -54,7 +54,7 @@ async fn main() -> anyhow::Result<()> {
     let server = axum_server::bind_rustls(
         addr,
         RustlsConfig::from_config(Arc::new(keypair.to_rpk_rustls_server_config())),
-    );
+    )?;
 
     let app = Router::new().route("/", get(|| async { "Hello, world!" }));
     server.serve(app.into_make_service()).await?;
--- a/pkarr/examples/http-serve.rs
+++ b/pkarr/examples/http-serve.rs
@@ -3,11 +3,9 @@
 //! This server will _not_ be accessible from other networks
 //! unless the provided IP is public and the port number is forwarded.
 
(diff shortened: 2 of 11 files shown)


pkarr/examples/http-get.rs
```diff
@@ -2,7 +2,6 @@
 
 use reqwest::Method;
 use tracing::Level;
-use tracing_subscriber;
```
SHAcollision, line 5: format/clippy fix

pkarr/examples/http-serve.rs
```diff
@@ -3,11 +3,9 @@
 //! This server will _not_ be accessible from other networks
 //! unless the provided IP is public and the port number is forwarded.
 
-use tracing::Level;
-use tracing_subscriber;
```
SHAcollision, line 7: format/clippy fix

relay/src/lib.rs
```diff
@@ -157,6 +157,7 @@ impl Relay {
         let client = config.pkarr.build()?;
 
         let listener = TcpListener::bind(SocketAddr::from(([0, 0, 0, 0], config.http_port)))?;
+        listener.set_nonblocking(true)?;
```
SHAcollision, line 162: Would not pass tests without this. Unsure if this is the correct solution.

afterburn: @SHAcollision I looked into this a little bit and believe this is the correct solution. In axum-server 0.7.x, they called `set_nonblocking(true)` internally before passing the listener to tokio. In 0.8.x, they removed that call, so it looks like we need to do it ourselves, which you did.

See [0.7.x docs](https://github.com/programatik29/axum-server/blob/v0.7.1/src/server.rs#L250) and [0.8.x docs](https://github.com/programatik29/axum-server/blob/v0.8.0/src/server.rs#L363) + the `from_tcp` method.

There's actually an open issue about this, their own example is missing the call too: https://github.com/programatik29/axum-server/issues/181

I think we could add a comment here explaining the reasoning for adding this line.

afterburn (changes requested):
Requested a small change

afterburn (approved):
👍

PR #201, opened 2026-02-03
feat!: Bump `simple-dns` dependency

ok300:
This PR updates the `simple-dns` dependency from `0.9.3` (~1y old) to the latest `0.11.2`.

The changelog is available here: https://github.com/balliegojr/simple-dns/blob/main/simple-dns/CHANGELOG.md

There were some breaking changes in the new `simple-dns`, like the use of parsed `SVCParam`s as an enum. The `pkarr` loops and callers were updated to reflect that.

A trickier part was the `Endpoint` behind the `extra` feature flag. It used to keep a copy of all params as

```diff
-    params: BTreeMap<u16, Box<[u8]>>,
+    params: BTreeMap<u16, SVCParam<'a>>,
```

but the newer `simple-dns` 1) doesn't expose the underlying bytes for the params anymore, and 2) instead exposes the params as parsed `SVCParam<'a>`.

### DIFF
--- a/bindings/js/src/dns/svcb/output.rs
+++ b/bindings/js/src/dns/svcb/output.rs
@@ -54,10 +54,7 @@ fn parse_param_value_from_param(param: &SVCParam) -> String {
         SVCParam::Port(port) => port.to_string(),
         SVCParam::Ipv4Hint(ips) => ips
             .iter()
-            .map(|&ip| {
-                let bytes = ip.to_be_bytes();
-                format!("{}.{}.{}.{}", bytes[0], bytes[1], bytes[2], bytes[3])
-            })
+            .map(|&ip| std::net::Ipv4Addr::from(ip).to_string())
             .collect::<Vec<_>>()
             .join(","),
         SVCParam::Ipv6Hint(ips) => ips
--- a/bindings/js/src/dns/svcb/output.rs
+++ b/bindings/js/src/dns/svcb/output.rs
@@ -2,17 +2,18 @@
 
 use super::constants::*;
 use super::utils::bytes_to_hex;
-use simple_dns::rdata::SVCB;
+use simple_dns::rdata::{SVCParam, SVCB};
 use wasm_bindgen::JsValue;
 
 /// Convert SVCB parameters to JavaScript object with descriptive keys and parsed values
 pub fn to_js_object(svcb: &SVCB) -> Result<js_sys::Object, JsValue> {
     let params_obj = js_sys::Object::new();
 
-    for (key, value) in svcb.iter_params() {
+    for param in svcb.iter_params() {
+        let key = param.key_code();
         if is_known_param(key) {
             let key_name = param_key_to_name(key);
-            let parsed_value = parse_param_value(key, value);
+            let parsed_value = parse_param_value_from_param(param);
             js_sys::Reflect::set(
                 &params_obj,
                 &JsValue::from_str(key_name),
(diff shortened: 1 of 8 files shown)


bindings/js/src/dns/svcb/output.rs
```diff
@@ -21,7 +22,10 @@ pub fn to_js_object(svcb: &SVCB) -> Result<js_sys::Object, JsValue> {
         } else {
             // For unknown parameters, use the format "param{number}" with hex values
             let unknown_key = format!("param{}", key);
-            let hex_value = bytes_to_hex(value);
+            let hex_value = match param {
+                SVCParam::Unknown(_, data) => bytes_to_hex(data.as_ref()),
+                _ => bytes_to_hex(&[]), // This shouldn't happen for truly unknown params
```
dzdidi: Not sure if this should just be silently swallowed ...

bindings/js/src/dns/svcb/output.rs
```diff
@@ -33,73 +37,36 @@ pub fn to_js_object(svcb: &SVCB) -> Result<js_sys::Object, JsValue> {
...
+            .join(","),
+        SVCParam::Port(port) => port.to_string(),
+        SVCParam::Ipv4Hint(ips) => ips
+            .iter()
+            .map(|&ip| {
+                let bytes = ip.to_be_bytes();
+                format!("{}.{}.{}.{}", bytes[0], bytes[1], bytes[2], bytes[3])
+            })
+            .collect::<Vec<_>>()
+            .join(","),
```
dzdidi, line 59:
```suggestion
        SVCParam::Ipv4Hint(ips) => ips
            .iter()
            .map(|&ip| std::net::Ipv4Addr::from(ip).to_string())
            .collect::<Vec<_>>()
            .join(","),
```

pkarr/src/extra/endpoints/endpoint.rs
```diff
@@ -12,23 +12,27 @@ use std::{
...
 /// An alternative Endpoint for a `qname`, from either [RData::SVCB] or [RData::HTTPS] dns records
-pub struct Endpoint {
+pub struct Endpoint<'a> {
     target: String,
     public_key: PublicKey,
     port: u16,
     /// SocketAddrs from the [SignedPacket]
     addrs: Vec<IpAddr>,
-    params: BTreeMap<u16, Box<[u8]>>,
+    params: BTreeMap<u16, SVCParam<'a>>,
```
dzdidi: Maybe just make it  `<'static>` ?
```suggestion
pub struct Endpoint {
    target: String,
    public_key: PublicKey,
    port: u16,
    /// SocketAddrs from the [SignedPacket]
    addrs: Vec<IpAddr>,
    params: BTreeMap<u16, SVCParam<'static>>,
```

Then you would not need changes at pkarr/src/extra/endpoints/mod.rs 

As you already calling to `into_owned`  inside of `parse` https://github.com/pubky/pkarr/pull/201/changes#diff-585d7cf91b08860b8fcdd559bde5e65f78f09fe14ec416c0f0b4a2c1306fe9d7R92

dzdidi (changes requested):

bindings/js/src/dns/svcb/output.rs
ok300: Throwing error instead: 009548582952bd040ae0009b61787175a6009ee2

pkarr/src/extra/endpoints/endpoint.rs
ok300: Done: f4fa1728cbbd41c30f108ed59d8b4db3f6e3e9aa

dzdidi (approved):
Code wise LGTM. Unfortunately can not speak beyond that so would wait for 2nd approval

SeverinAlexB:
Added `feat!` to the title to comply with conventional commits. Important: `!` indicates this is a breaking change.

I think we need to release this as a release candidate first to make sure it will not have unintended consequences down the road.

ok300:
Superseded by #207

PR #203, opened 2026-02-05
fix: replace relay.pkarr.org with pkarr.pubky.app

gcomte:
The pkarr.org domain is not controlled by us.

### DIFF
--- a/relays.txt
+++ b/relays.txt
@@ -1,2 +1,2 @@
-https://relay.pkarr.org
+https://pkarr.pubky.app
 https://pkarr.pubky.org
--- a/relays.txt
+++ b/relays.txt
@@ -1,2 +1,3 @@
 https://pkarr.pubky.app
 https://pkarr.pubky.org
+https://relay.pkarr.org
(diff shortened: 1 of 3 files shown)


relays.txt
```diff
@@ -1,2 +1,2 @@
-https://relay.pkarr.org
+https://pkarr.pubky.app
 https://pkarr.pubky.org
```
SeverinAlexB, line 2: You can keep `https://relay.pkarr.org` in this file, just move it down. This is just a list with known relays so it doesn't hurt.

SeverinAlexB (changes requested):
You are correct in making this PR but wait with this change.

`https://pkarr.pubky.app` is hosted in the Google Cloud and mainline has issues there. So we first need to wait for @afterburn to fix the relay in the Google Cloud before we add it as the default relay.

relays.txt
gcomte: 0ec2920308a1473895f356b5c8b1fc0c6ffa718c :heavy_check_mark:

gcomte:
Let's leave it as a draft as long as `https://pkarr.pubky.app` is not stable yet.

SeverinAlexB (approved):
FYI: `cargo deny` failed due to a security advisory issue. This is unrelated to this PR and needs attention in a new PR.

PR #205, opened 2026-02-05
chore: Github Action Improvements

SeverinAlexB:
## Summary                                                                                                                                                                        
                                                                                                                                                                                    
Overhauled GitHub Actions workflows for faster CI (from 9m to 1m), better reliability, and easier releases.                                                                                       
 
## Changes                                                                                                                                                                        
                                                                                                                                                                                
### CI Improvements (`rust.yaml`, `wasm.yaml`, `js-binding.yaml`)
- Replaced deprecated actions with modern alternatives
- Parallelized rust.yaml into 7 independent jobs
- Added `Swatinem/rust-cache@v2` everywhere
- Added concurrency control (cancels redundant runs)
- Stricter linting with `-D warnings`
- Added MSRV (1.85) testing

### Release Workflow (`build-artifacts.yaml`)
- Replaced `cross` tool with native GitHub runners
- Auto-creates GitHub Release on tag push (`v*`)
- Kept musl for static Linux binaries

| Target | Runner |
|--------|--------|
| linux-amd64 (musl) | `ubuntu-latest` |
| linux-arm64 (musl) | `ubuntu-latest-arm64` |
| osx-amd64 | `macos-latest` |
| osx-arm64 | `macos-latest` |
| windows-amd64 | `windows-latest` |

### Bug Fixes
- Fixed rustdoc warning in `bindings/js/src/dns/svcb/output.rs`
- Fixed clippy warning in `relay/src/handlers.rs`


---

The `cargo deny` test fail will resolve with #204 . We can therefore merge this PR despite it failing.


.github/workflows/build-artifacts.yaml
```diff
@@ -0,0 +1,109 @@
...
+          tar -czvf "$FOLDER.tar.gz" "$FOLDER"
+
+      - name: Package (Windows)
+        if: runner.os == 'Windows'
+        shell: bash
+        run: |
+          FOLDER="pkarr-relay-v${{ steps.version.outputs.version }}-${{ matrix.name }}"
+          mkdir -p "$FOLDER"
+          cp target/${{ matrix.target }}/release/pkarr-relay.exe "$FOLDER/"
+          tar -czvf "$FOLDER.tar.gz" "$FOLDER"
```
86667: This look like it could be more succint

.github/workflows/ci-wasm.yaml
```diff
@@ -2,40 +2,61 @@ name: WASM Tests
...
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Install wasm-pack
```
86667, line 30: we do this as a cargo install in pubky-core. a quick AI search suggests `taiki-e/install-action@wasm-pack` is better than both appraoches for its speed and caching

.github/workflows/build-artifacts.yaml
SeverinAlexB: How? From my experience, making GH Actions DRY is very ugly and increases complexity. Do you see something I don't?

.github/workflows/ci-wasm.yaml
SeverinAlexB: Not sure we need to make this perfect perfect. I can have a look but is it worth the time spent?

.github/workflows/build-artifacts.yaml
86667: Im seeing I think 3 identical lines, its only the `.exe` suffix added in the windows branch as far as i can tell. 
Its fine, just looks like it could be cleaner

.github/workflows/ci-wasm.yaml
86667: Agreed. Feels like a good opportunity to choose a standard though. it seems like an uncomplication decision to make

.github/workflows/build-artifacts.yaml
SeverinAlexB: In regular code, I would agree with you. In Github actions, removing repetition is a bitch.

86667: Couldnt you eg use a wildcard instead of specifying the path - `cp target/${{ matrix.target }}/release/pkarr-relay* "$FOLDER/"`
I may be oversimplifying. conditional branches for windows for this simple zip task just feels wrong

PR #214, opened 2026-02-11
fix: tests on MacOS

SeverinAlexB:
## Summary
- Bind all test DHT nodes and clients to `127.0.0.1` instead of `0.0.0.0` to fix 46 test failures on macOS
- Replace deprecated `Testnet::new_async(N)` / `Testnet::new(N)` with `Testnet::builder(N).build()` (defaults to localhost)
- Add `.bind_address(Ipv4Addr::LOCALHOST)` to all test `DhtBuilder` configs

## Context

On newer macOS, binding UDP sockets to `0.0.0.0` causes DHT nodes to become unreachable — the OS may route UDP responses through a different interface than loopback, resulting in `NoClosestNodes` and `UnexpectedResponses` errors. This caused 46 out of ~96 tests to fail.

Mainline 6.1.1 added `Testnet::builder(N)` which defaults `bind_address` to `127.0.0.1` (vs `0.0.0.0` in the legacy `new()` / `new_async()` constructors). This PR adopts that pattern across all pkarr tests, matching what mainline's own test suite does.

## Changes

- **`pkarr/src/client/tests.rs`**: Replaced all `Testnet::new_async()` calls with `Testnet::builder().build()`, added `.dht(|b| b.bind_address(Ipv4Addr::LOCALHOST))` to the shared `builder()` helper and standalone client builders
- **`pkarr/src/extra/endpoints/mod.rs`**: Same pattern for all 5 endpoint tests
- **`relay/src/lib.rs`**: Added `bind_address(LOCALHOST)` to `run_test()` and `run_testnet()`, replaced `Testnet::new()` with `Testnet::builder().build()`

See https://github.com/pubky/mainline/pull/82

### DIFF
--- a/relay/src/lib.rs
+++ b/relay/src/lib.rs
@@ -240,7 +240,7 @@ impl Relay {
     /// Homeserver uses LMDB, opening which is marked [unsafe](https://docs.rs/heed/latest/heed/struct.EnvOpenOptions.html#safety-1),
     /// because the possible Undefined Behavior (UB) if the lock file is broken.
     pub async unsafe fn run_testnet() -> anyhow::Result<Self> {
-        let testnet = pkarr::mainline::Testnet::new_async(10).await?;
+        let testnet = pkarr::mainline::Testnet::builder(10).build()?;
 
         // Leaking the testnet to avoid dropping and shutting them down.
         for node in testnet.nodes {
--- a/relay/src/lib.rs
+++ b/relay/src/lib.rs
@@ -205,7 +205,7 @@ impl Relay {
     }
 
     /// Run an ephemeral Pkarr relay on a random port number for testing purposes.
-    ///
+    /// Binds to `127.0.0.1`.
     /// # Arguments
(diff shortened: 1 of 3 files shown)


relay/src/lib.rs
```diff
@@ -236,7 +240,7 @@ impl Relay {
     /// Homeserver uses LMDB, opening which is marked [unsafe](https://docs.rs/heed/latest/heed/struct.EnvOpenOptions.html#safety-1),
     /// because the possible Undefined Behavior (UB) if the lock file is broken.
     pub async unsafe fn run_testnet() -> anyhow::Result<Self> {
-        let testnet = pkarr::mainline::Testnet::new(10)?;
+        let testnet = pkarr::mainline::Testnet::new_async(10).await?;
```
afterburn: nit: probably should use `Testnet::builder(10).build()?` here for consistency, we use it in the other tests

afterburn (approved):
@SeverinAlexB lgtm, one small nit but feel free to ignore

PR #234, opened 2026-04-29
build: disable default features for dependencies

andrei-21:
Disable unnecessary default features and remove unused relay dependencies to reduce the resolved dependency graph.

Use reqwest `rustls-no-provider` only where TLS integration is needed.


SeverinAlexB (approved):
LGTM

PR #236, opened 2026-05-05
refactor(relay): split client IP normalization and rate limiting

andrei-21:
Prepare outbound DHT request rate limiting to reuse the same
real-IP extraction path as HTTP rate limiting.

### DIFF
--- a/relay/src/rate_limiting.rs
+++ b/relay/src/rate_limiting.rs
@@ -1,14 +1,10 @@
 use std::{net::IpAddr, sync::Arc, time::Duration};
 
+use crate::real_ip::RealIpKeyExtractor;
 use axum::Router;
 use governor::middleware::StateInformationMiddleware;
 use serde::{Deserialize, Serialize};
-
-use tower_governor::{
-    governor::{GovernorConfig, GovernorConfigBuilder},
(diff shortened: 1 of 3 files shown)


relay/src/rate_limiting.rs
```diff
@@ -43,80 +39,49 @@ impl Default for RateLimiterConfig {
...
+
+        // The governor needs a background task for garbage collection (to clear expired records)
+        let gc_interval = Duration::from_secs(60);
+
+        let governor_limiter = governor_middleware.limiter().clone();
+        std::thread::spawn(move || loop {
+            std::thread::sleep(gc_interval);
+            tracing::debug!("rate limiting storage size: {}", governor_limiter.len());
+            governor_limiter.retain_recent();
+        });
```
SeverinAlexB, line 70: NIT, already existing so not cause by you: thread is quite heavy. How about `tokio::spawn`?

SeverinAlexB (approved):
Nice. One NIT. No need to change if you don't feel like it. It's rather cosmetic.

relay/src/rate_limiting.rs
andrei-21: Makes sense, I will take a note to address it later.

PR #239, opened 2026-05-08
feat(relay): add resolve policies

andrei-21:


pkarr/src/client.rs
```diff
@@ -56,6 +57,18 @@ pub(crate) struct Inner {
...
+/// Controls which source a resolve request is allowed to use.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Deserialize)]
+pub enum ResolvePolicy {
+    /// Return only a locally cached packet.
+    LocalOrRelayCacheOnly,
+    /// Return cached data first, falling back to the network on cache miss.
+    #[default]
+    CacheFirst,
+    /// Query only the DHT network and ignore the local cache.
+    DhtNetworkOnly,
```
dzdidi: Looks like there is no way to force update cache, so if there is something "broken" stored in cache the only way is TTL.

Maybe it is worth changing `DhtNetworkOnly` (or adding new policy) that would read from DHT, update cache and return newly read value?

andrei-21: The client of the relay should not manipulate the cache of the relay directly.
It is up to the relay to decide when to update the cache, however the cache will try it's best to have items as complete and as up-to-date as possible.

pkarr/src/resolve_policy.rs
```diff
@@ -0,0 +1,31 @@
+//! Resolve policies.
+
+use serde::Deserialize;
+
+/// controls whether resolution may use only cached packets,
```
SeverinAlexB:
```suggestion
/// Controls whether resolution may use only cached packets,
```

pkarr/src/resolve_policy.rs
```diff
@@ -0,0 +1,31 @@
...
+/// prefer cached packets before querying the DHT, or bypass the cache and query
+/// the DHT network directly.
+///
+/// | Policy | Cache behavior | DHT behavior | Typical use |
+/// | --- | --- | --- | --- |
+/// | [`ResolvePolicy::LocalOrRelayCacheOnly`] | Returns a cached packet, even if expired | Never queried | Fast local reads and republishing |
```
SeverinAlexB: Is this the correct definition? "Never queried" is missing the what is queried. "Never queries the DHT" is more accurate as it may query the relay.

"Fast local reads and republishing" Local is missleading because it could also query the relay.

What do you think?

pkarr/src/resolve_policy.rs
```diff
@@ -0,0 +1,31 @@
...
+/// | [`ResolvePolicy::CacheFirst`] | Returns a cached packet only while it is fresh | Queries the first responder on cache miss or expiry | Normal application resolution while respecting TTLs |
```
SeverinAlexB: Wording proposals:
"Returns a cached packet only while it is fresh" -> Hard to read. What about "Uses the cache if the packet is available and if it is within the TTL."?
"Queries the first responder on cache miss or expiry" -> I know what the first responder is but I feel like a pkarr library user will likely not. How about: "Queries the DHT on a cache miss. Uses the first DHT response to guarantee a speedy response. May return an outdated packet in edge cases."

pkarr/src/resolve_policy.rs
```diff
@@ -0,0 +1,31 @@
...
+/// | [`ResolvePolicy::DhtNetworkOnly`] | Ignores cached packets for lookup, but updates the cache after success | Queries relevant DHT nodes for the newest packet | Recovering after stale sequence errors |
```
SeverinAlexB: How about this? I am trying to make it as easy to understand as possible.

```suggestion
/// | [`ResolvePolicy::DhtNetworkOnly`] | Ignores cached packets for the lookup, but updates the cache after retrieval | Queries all relevant DHT nodes for the newest packet. Slow but accurate. | When you need the absolute most recent packet, for example for recovering after stale sequence errors |
```

pkarr/src/resolve_policy.rs
```diff
@@ -0,0 +1,31 @@
...
+    /// from the first responding DHT node otherwise.
+    ///
+    /// This is guaranteed to return only non-expired packets.
+    CacheFirst,
+
+    /// Query all relevant DHT nodes and update the cache.
+    ///
+    /// This is guaranteed to return the newest packet.
+    DhtNetworkOnly,
+}
```
SeverinAlexB, line 33: See my inputs about the comments above. They should likely be applied here too.

design/relays.md
```diff
@@ -63,13 +63,16 @@ Content-Type: application/pkarr.org/relays#payload
...
 Last-Modified: Fri, 18 Oct 2024 13:24:21 GMT
+Memento-Datetime: Fri, 8 May 2026 04:21:21 GMT
 
 <body>
 ```
 
 `Cache-Control` header would help browsers reduce their reliance on the relay, the `max-age` should be set to be the minimum `ttl` in the resource records in the packet or some minimum ttl chosen by the relay.
 `If-Modified-Since` can be sent by the client to avoid downloading packets they already have, when the relay responds with `304 Not Modified`.
 
+`Memento-Datetime` will tell when the last time the packet was observed on the DHT network.
```
SeverinAlexB: Some grammar

```suggestion
`Memento-Datetime` will tell you when the packet was last observed on the DHT network.
```

relay/src/handlers.rs
```diff
@@ -26,50 +24,89 @@ pub async fn put(
...
 }
 
 pub async fn get(
     State(state): State<AppState>,
     Path(public_key): Path<String>,
+    Query(query): Query<GetQuery>,
     request_headers: HeaderMap,
 ) -> Result<impl IntoResponse, Error> {
     let public_key = PublicKey::try_from(public_key.as_str())
         .map_err(|error| Error::new(StatusCode::BAD_REQUEST, Some(error)))?;
```
SeverinAlexB: Already existing but you could safe the whole PublicKey transformation line/error by making axum handle the input validation.

```suggestion
Path(public_key): Path<PublicKey>,
    Query(query): Query<GetQuery>,
request_headers: HeaderMap,
) -> Result<impl IntoResponse, Error> {
```

relay/src/handlers.rs
```diff
@@ -26,50 +24,89 @@ pub async fn put(
...
+        })?
+        .map(|s| s.parse::<u64>())
+        .transpose()
+        .map_err(|_| {
+            Error::new(
+                StatusCode::BAD_REQUEST,
+                Some("Invalid IF_MATCH header value"),
+            )
+        })?
         .map(Timestamp::from);
```
SeverinAlexB: You can make this a lot cleaner by defining the header in an axum extractor:

```rust
/// Compare-And-Swap token parsed from the `If-Match` request header.
///
/// The pkarr relay reuses `If-Match` to carry a u64 sequence number rather
/// than an ETag list, so it cannot be modeled with `headers::IfMatch`.
///
/// Returns `Self(None)` when the header is absent and a `400 Bad Request`
/// when the header is present but not a valid u64.
#[derive(Debug)]
pub(crate) struct IfMatchCas(pub Option<Timestamp>);

impl<S: Send + Sync> FromRequestParts<S> for IfMatchCas {
    type Rejection = Error;

    async fn from_request_parts(parts: &mut Parts, _state: &S) -> Result<Self, Error> {
        let Some(value) = parts.headers.get(header::IF_MATCH) else {
            return Ok(Self(None));
        };
        let timestamp = value
            .to_str()
            .ok()
            .and_then(|s| s.parse::<u64>().ok())
            .map(Timestamp::from)
            .ok_or_else(|| {
                Error::new(
                    StatusCode::BAD_REQUEST,
                    Some("Invalid IF_MATCH header value"),
                )
            })?;
        Ok(Self(Some(timestamp)))
    }
}
```

And then use the extractor in the handler

```rust
pub async fn put(
    State(state): State<AppState>,
    Path(public_key): Path<PublicKey>,
    IfMatchCas(cas): IfMatchCas,
    body: Bytes,
```

This is AI generated so take it with a grain of salt but the extractor is the way to go IMO.

Another note: We use a non standard compliant `IF_MATCH` header here. No need to do anything but in the future, we might want to be more standard compliant.

relay/src/handlers.rs
```diff
@@ -26,50 +24,89 @@ pub async fn put(
...
+
+    let cas = cas.map(|t| t.as_u64() as i64);
+    // TODO: Ratelimit.
+    state.dht.put_mutable((&signed_packet).into(), cas).await?;
+    update_cache(&state, &key, &signed_packet);
 
     Ok(StatusCode::NO_CONTENT)
 }
 
 pub async fn get(
```
SeverinAlexB, line 44: This method is getting really long. This is not caused by you but it intensifies with your additions.
https://refactoring.guru/smells/long-method

relay/src/handlers.rs
```diff
@@ -120,14 +165,19 @@ pub async fn get(
     }
 }
 
+#[derive(Debug, Deserialize)]
+pub(crate) struct GetQuery {
+    policy: Option<ResolvePolicy>,
+}
```
SeverinAlexB, line 100: Nice

SeverinAlexB (changes requested):
The change looks good functionally.

The existing code quality is not good so a lot of my comments are directed towards this and not your additions. I recommend you adopt the [Boy Scount Principle](https://medium.com/@codechuckle/code-quality-unleashed-boy-scout-principle-6b845e831424) to improve the existing code PR after PR.

If you have any questions let me know.

andrei-21:
The Boy Scout principle is in my blood, fully agree, however sometimes you need to stop improving to keep the scope of the changes small.
Your suggestions are very reasonable, I applied your minor suggestions, added extractors to handle requests and response parts, it indeed makes the code nicer. 
I will address your concerns about the wording in docs.

andrei-21:
I reworked `ResolvePolicy` documentation to make simpler, cleaner and more precise.

pkarr/src/resolve_policy.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// This is guaranteed to return only packets that are not expired.
+    /// Useful for normal application resolution while respecting TTLs.
+    CacheFirst,
+
+    /// Query all relevant DHT nodes for the most recent packet observed and
+    /// update the cache after retrieval.
+    /// This is slower, but more accurate.
+    ///
+    /// This returns the most recent packet found during the query, but does not
+    /// guarantee global consistency.
```
SeverinAlexB: The other comments look good but here I have a question: What do you mean with "does not guarantee global consistency"?

If we query all 20 DHT nodes, we are like 99.99% sure we have the most recent packet. Are you trying to communicate that even then, it's not 100% guaranteed?

relay/src/handlers.rs
```diff
@@ -1,133 +1,112 @@
...
-        if let Some(condition_http_date) = request_headers
-            .get(header::IF_MODIFIED_SINCE)
-            .and_then(|h| h.to_str().ok())
-            .and_then(|s| HttpDate::from_str(s).ok())
-        {
-            let entry_http_date: HttpDate = signed_packet.timestamp().into();
-
-            if condition_http_date >= entry_http_date {
-                *response.status_mut() = StatusCode::NOT_MODIFIED;
+    Path(PublicKeyParam(public_key)): Path<PublicKeyParam>,
```
SeverinAlexB, line 46: Instead of creating this `PublicKeyParam` wrapper, can't you just put `PublicKey` in here and it works already? I might be wrong but I think PublicKey already implements the serde deserializer.

relay/src/handlers.rs
```diff
@@ -1,133 +1,112 @@
...
+                &public_key,
+                Mode::MostRecent,
+                at_least_that_recent,
+            )
+            .await
+        }
+    };
+
+    if let Some(signed_packet) = signed_packet {
+        update_cache(&state, &key, &signed_packet);
```
SeverinAlexB: Looking at the code: If the `policy` is `LocalOrRelayCacheOnly` and the packet was found in the cache, aren't you refreshing the cache here which I guess you shouldn't because it is coming from the cache?

I might be wrong but from reading the high level code, I see that a cached packet is turned into the signed_packet and then the cache is updated which feels wrong.

SeverinAlexB (changes requested):
Great change. A few minor things and then we are good to go.

> however sometimes you need to stop improving to keep the scope of the changes small.

I agree. It's also not always clear how much cleaning up is reasonable.

relay/src/handlers.rs
andrei-21: It does not implement, and I am not sure if it makes sense to add it, since it is not clear how to deserialize—from a string or from bytes.

relay/src/handlers.rs
andrei-21: No, but I refactored the method to be more explicit and renamed to `update_cache_if_needed()`.

pkarr/src/resolve_policy.rs
andrei-21: Probably the users do not need to be overwhelmed with these details, I returned the stronger wording.

SeverinAlexB (approved):
Let's go

Issue #241, opened 2026-05-13
[Breaking] v5.0.4 adds a breaking change by bumping the simple-dns dependency


SeverinAlexB:
The `simple-dns` dependency bump to 0.11.2 was intended for v6 but accidentally released with pkarr [`v5.0.4`](https://github.com/pubky/pkarr/releases/tag/v5.0.4) 2 months ago. This is a breaking change that requires a major version bump and not a patch.

Sneaked in with #217

SeverinAlexB:
`v5.0.4` and `v5.0.5` have been yanked. Use the newly released `v6.0.0` if you want to keep the breaking change.

PR #243, opened 2026-05-13
chore: v6.0.0

SeverinAlexB:
Version 6.0.0 which replaces v5.0.5 due to #241 .

### DIFF
--- a/relay/Cargo.toml
+++ b/relay/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pkarr-relay"
-version = "0.11.6"
+version = "0.11.7"
 rust-version = "1.85"
 authors = [
     "Nuh <nuh@nuh.dev>",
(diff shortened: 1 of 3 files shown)


relay/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pkarr-relay"
-version = "0.11.6"
+version = "0.11.7"
```
86667: Lets do a minor bump here since it bumped pkarr with a breakign change

86667 (changes requested):

relay/Cargo.toml
SeverinAlexB: The relay doesnt change it's http api. For the relay, the pkarr bump is a patch as far as I can see. Correct me if I am missing something

86667: The http api doesnt change, but Relay does expose its internal `pkarr: pkarr::ClientBuilder`, so technically i think we should bump.

SeverinAlexB: Agree. Done

86667 (approved):

PR #244, opened 2026-05-14
feat(relay)!: rate limit user DHT operations

andrei-21:
Add a separate user-initiated DHT rate limiter for relay HTTP
PUT requests and GET requests that query the DHT. Preserve cache-only
and fresh CacheFirst reads without consuming that quota.

BREAKING CHANGE: RateLimiterConfig now requires user_dht_per_second
and user_dht_burst_size fields when constructed with a Rust struct
literal.

### DIFF
--- a/relay/src/config.example.toml
+++ b/relay/src/config.example.toml
@@ -37,11 +37,11 @@ maximum_ttl =  86400
 # Ip rate limiting by setting the trusted headers to 
 # a random IP address on every request.
 behind_proxy = false
-# Maximum number of requests per second.
-burst_size = 10
-# Number of seconds after which one request of the quota is replenished.
-per_second = 2
-# Maximum number of DHT operations (initiated by user's request) per second.
-user_dht_burst_size = 5
-# Number of seconds after which one user DHT operation quota is replenished.
-user_dht_per_second = 2
+# HTTP request quota per IP address.
+quota = "2r/s"
+# Optional HTTP request burst size. Defaults to the quota rate when omitted.
+burst = 10
+# DHT operation quota per IP address for operations initiated by user's request.
+user_dht_quota = "2r/s"
+# Optional user DHT operation burst size. Defaults to the quota rate when omitted.
+user_dht_burst = 10
--- a/relay/src/config.example.toml
+++ b/relay/src/config.example.toml
@@ -38,10 +38,17 @@ maximum_ttl =  86400
 # a random IP address on every request.
 behind_proxy = false
 # HTTP request quota per IP address.
+# Always applied before request handling to every HTTP request.
+# This includes GET, PUT, the index route, cache hits, and cache-only requests.
 quota = "2r/s"
 # Optional HTTP request burst size. Defaults to the quota rate when omitted.
 burst = 10
-# DHT operation quota per IP address for operations initiated by user's request.
+# User-initiated DHT operation quota per IP address.
+# This is separate from the HTTP quota above and is consumed only when a
+# request makes the relay contact the DHT: PUT publishes, GET cache misses,
+# expired CacheFirst entries, and DhtNetworkOnly requests. Cache hits and
+# LocalOrRelayCacheOnly requests do not consume this quota.
 user_dht_quota = "2r/s"
-# Optional user DHT operation burst size. Defaults to the quota rate when omitted.
+# Optional user-initiated DHT operation burst size.
+# Defaults to the DHT quota rate when omitted.
 user_dht_burst = 10
--- a/relay/src/config.example.toml
+++ b/relay/src/config.example.toml
@@ -41,3 +41,7 @@ behind_proxy = false
 burst_size = 10
 # Number of seconds after which one request of the quota is replenished.
 per_second = 2
+# Maximum number of DHT operations (initiated by user's request) per second.
+user_dht_burst_size = 5
+# Number of seconds after which one user DHT operation quota is replenished.
+user_dht_per_second = 2
--- a/relay/src/lib.rs
+++ b/relay/src/lib.rs
@@ -11,6 +11,7 @@ mod config;
 mod error;
 mod extractors;
 mod handlers;
+mod quota_config;
 mod rate_limiting;
 mod real_ip;
 mod response;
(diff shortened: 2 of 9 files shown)


relay/src/rate_limiting.rs
```diff
@@ -26,6 +26,47 @@ pub struct RateLimiterConfig {
...
+    /// When omitted from configuration, this defaults to [`Self::per_second`].
+    ///
+    /// **The interval must not be zero.**
+    pub user_dht_per_second: u64,
+    /// Set quota size for DHT operations initiated by HTTP user requests.
+    ///
+    /// When omitted from configuration, this defaults to [`Self::burst_size`].
+    ///
+    /// **The burst_size must not be zero.**
+    pub user_dht_burst_size: u32,
```
SeverinAlexB: Is per second a good time range? 1 DHT query per second is already plenty IMO. Would 10 per minute + 60 burst not be better to prevent long term abuse? *Asking ignorantly

We do have a rate limiter layer in the homeserver that you could copy. Too complex for this case but might be an inspiration.
- https://github.com/pubky/pubky-core/blob/main/pubky-homeserver/src/client_server/layers/rate_limiter/request_rate_limit.rs
- https://github.com/pubky/pubky-core/blob/main/pubky-homeserver/src/data_directory/quota_config/path_limit.rs#L18

It allows you to configure `10r/s` or `50r/h`. At the same time, make sure to push back if you think this is too much

relay/src/lib.rs
```diff
@@ -144,18 +144,22 @@ impl Relay {
...
+                config.pkarr.dht(|builder| {
+                    builder.server_settings(pkarr::mainline::ServerSettings {
+                        filter: Box::new(rate_limiter.clone()),
+                        ..Default::default()
+                    })
+                });
+                (Some(rate_limiter), Some(user_dht_rate_limiter))
+            }
+            None => (None, None),
+        };
```
SeverinAlexB, line 169: This should likely be an axum layer?

SeverinAlexB (changes requested):
Make sure to push back on my comments if you think they are unreasonable. I am off until Monday lunch so this was a very fast review from my side so I don't slow you down.

relay/src/lib.rs
andrei-21: Not sure I understand what you mean.
`rate_limiter` is already a layer, `user_dht_rate_limiter` cannot be a layer, because it is not applied uniformly to all requests, but only to requests causing requests to DHT nodes.

relay/src/rate_limiting.rs
andrei-21: Hm, let's run the numbers. If we allow up to 10 requests per second, our homeserver will be able to republish records for only 600 keys (given we want to republish every hour). That does not look like it will be enough now or in the near future. However, we can whitelist our homeserver, but we need to add this functionality first.

Also, in normal operations, requests will be forwarded to different DHT nodes if keys are different. We can rate-limit by keys also, but it can still be abused if an attacker generates different keys that fall into the same buckets.

What are your thoughts on this?

SeverinAlexB: With Kevin, we were discussing this problem and out came the following proposal:
- The homeserver gets it's own private relay with no rate limiting so it can publish as much as it wants.
- Public relays for web applications do have a relatively strict limit.

This would prevent us from needing a white list. What do you think?

relay/src/lib.rs
SeverinAlexB: You are correct, it is already a layer. Let me rephrase it: This code looks really out of place and should be living close to the rate limiter IMO. @86667 Can you have a look here from a clean code POV?

relay/src/rate_limiting.rs
andrei-21: Yes, sounds reasonable. I will add an option to set the limits more flexible.

relay/src/lib.rs
86667: The concern which I image Sev has is that we have a `UserDhtRateLimiter` in `AppState` which provides an `is_limited()` method. Here we implement a function which checks if a rate limiter and ip exist, calls `is_limited()` if so and defines the error type to return on failure - behaviour which is specific to the `UserDhtRateLimiter`, yet it lives with the handler code.

I wonder if we could take it a step further and define some kind of `DhtOperator` which groups DHT operations with dht rate limiter together, so that all dht logic is in a single place and accessible via `AppState`?

andrei-21: Hm, not sure I understand, but if you refer to this line `builder.server_settings()` then I agree. However, I do not want to change it in this PR. The logic of applying users' HTTP rate limiter to requests from DHT nodes is so strange that it should be tackled separately.

Introducing `DhtOperator` also looks strange to me because the nature of DHT requests initiated by users is very different from the nature of requests from DHT nodes. One is a relay; the other is a DHT node.

SeverinAlexB: Thx @86667.
I just had a call with Andrei and agreed to tackle this in another PR.

relay/src/rate_limiting.rs
andrei-21: https://github.com/pubky/pkarr/pull/244/changes/00e70257dd186ca852c357c07a6777482a9aa588 now allows to set the limits in `Xr/s` format.

relay/src/config.example.toml
```diff
@@ -37,11 +37,11 @@ maximum_ttl =  86400
...
-# Number of seconds after which one user DHT operation quota is replenished.
-user_dht_per_second = 2
+# HTTP request quota per IP address.
+quota = "2r/s"
+# Optional HTTP request burst size. Defaults to the quota rate when omitted.
+burst = 10
+# DHT operation quota per IP address for operations initiated by user's request.
+user_dht_quota = "2r/s"
+# Optional user DHT operation burst size. Defaults to the quota rate when omitted.
+user_dht_burst = 10
```
SeverinAlexB, line 54: Do we need more specific naming/commenting here?

Is this the http `get` quota, or the `put` quota? Or are they combined? Are they used when they hit the cache?
Same for user_dht. An outsider would not know what this is

SeverinAlexB (changes requested):
One thing, looking good otherwise

relay/src/config.example.toml
andrei-21: Clarified.

SeverinAlexB (approved):
Let's go

PR #246, opened 2026-05-21
refactor(types): group protocol types in module

andrei-21:
Move key, signed packet, and resolve policy modules under
`pkarr::types` while preserving root-level re-exports.


andrei-21:
That is how `dht` module will look like: https://github.com/pubky/pkarr/pull/247.

pkarr/src/dht/mod.rs
```diff
@@ -0,0 +1,79 @@
+use ed25519_dalek::Signature;
+use mainline::MutableItem;
+use ntimestamp::Timestamp;
+
+use crate::{PublicKey, SignedPacket};
+
+impl From<&SignedPacket> for MutableItem {
```
86667: Could these live with `SignedPacket` code, we're tending towards a pattern of `mod.rs` files only for exports

pkarr/src/lib.rs
```diff
@@ -52,11 +46,8 @@ pub use simple_dns as dns;
 pub mod errors {
     //! Exported errors
 
-    #[cfg(feature = "keys")]
-    pub use super::keys::PublicKeyError;
-
-    #[cfg(feature = "signed_packet")]
-    pub use super::signed_packet::{SignedPacketBuildError, SignedPacketVerifyError};
+    #[cfg(any(feature = "keys", feature = "signed_packet"))]
```
86667: Youre feature-gating twice here. Once inside the module and once here.

Also why change to exporting all instead of being explicit?

andrei-21: I removed this line.
The point of exporting with `*` at the `lib.rs` level is to give modules autonomy over what they want to export and to avoid repeating exports at the `lib.rs` level.

pkarr/src/dht/mod.rs
andrei-21: It is an internal to `dht` module conversion. `types` should not depend on it.
I moved the code to `client.rs` where DHT client will appear in the following PRs.

86667 (approved):

PR #249, opened 2026-05-25
refactor(dht)!: hide mainline client behind wrapper

andrei-21:
BREAKING CHANGE: `ClientBuilder::dht` now configures `mainline::Config`
instead of `mainline::DhtBuilder`, `Client::dht` returns `DhtClient` instead
of `mainline::Dht`, and `ClientBlocking::dht` has been removed.

Introduce `DhtClient` as the internal DHT abstraction and move direct mainline Dht/AsyncDht usage behind `pkarr::dht`. The wrapper owns the AsyncDht, exposes publish, resolve, and info operations, and maps mainline publish failures into `pkarr::dht::PublishError`.

Switch client and relay configuration to pass `mainline::Config` instead of exposing `mainline::DhtBuilder` or `mainline::Dht`. This keeps pkarr's public DHT surface focused on pkarr types while still allowing callers to configure the underlying node.

Update relay GET handling to consume DHT response streams through the wrapper, return the first packet for cache-first reads, and continue processing later DHT responses in the background so newer packets can refresh the cache. Add tests for most-recent packet selection.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1480,12 +1480,11 @@ checksum = "112b39cec0b298b6c1999fee3e31427f74f676e4cb9879ed1a121b43661a4154"
 
 [[package]]
 name = "mainline"
-version = "6.1.2"
+version = "6.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


SeverinAlexB (approved):
LGTM. One validating question: Is the PR title correct, no breaking change?

andrei-21:
Right, I updated the message.

PR #250, opened 2026-05-26
feat(relays): expose single-relay client

andrei-21:
Add a public RelayClient for publishing and resolving SignedPackets
against one relay endpoint. Move single-relay HTTP behavior out of the
multi-relay client and map relay failures through a dedicated RelayError.

### DIFF
--- a/pkarr/src/client/relays.rs
+++ b/pkarr/src/client/relays.rs
@@ -5,28 +5,20 @@ use std::pin::Pin;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-use bytes::Bytes;
 use futures_buffered::FuturesUnorderedBounded;
 #[cfg(not(wasm_browser))]
 use futures_lite::Stream;
 use futures_lite::StreamExt;
 use ntimestamp::Timestamp;
-use reqwest::Method;
 use url::Url;
 
-use reqwest::{
-    header::{self, HeaderValue},
-    Client, StatusCode,
-};
-
 use super::{ConcurrencyError, PublishError, QueryError};
+use crate::relay_client::{RelayClient, RelayError};
 use crate::{PublicKey, SignedPacket};
 
 #[derive(Clone)]
 pub struct RelaysClient {
-    relays: Box<[Url]>,
-    http_client: Client,
-    timeout: Duration,
+    relays: Box<[RelayClient]>,
     pub(crate) inflight_publish: InflightPublishRequests,
 }
 
@@ -40,7 +32,7 @@ impl Debug for RelaysClient {
                 .relays
                 .as_ref()
                 .iter()
-                .map(|url| url.as_str())
+                .map(|relay| relay.base_url().as_str())
                 .collect::<Vec<_>>(),
         );
 
@@ -51,14 +43,16 @@ impl Debug for RelaysClient {
 impl RelaysClient {
     pub fn new(relays: Box<[Url]>, timeout: Duration) -> Self {
         let inflight_publish = InflightPublishRequests::new(relays.len());
+        let relays = relays
+            .into_vec()
+            .into_iter()
+            .map(|url| {
+                RelayClient::new(url, timeout).expect("Client building should be infallible")
+            })
+            .collect();
 
         Self {
             relays,
-            http_client: Client::builder()
-                .build()
-                .expect("Client building should be infallible"),
-
-            timeout,
             inflight_publish,
         }
     }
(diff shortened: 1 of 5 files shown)


pkarr/src/relay_client.rs
```diff
@@ -0,0 +1,647 @@
...
+        packet: &SignedPacket,
+        cas: Option<Timestamp>,
+    ) -> Result<(), RelayError> {
+        let url = self.build_url(&packet.public_key(), None);
+
+        let mut request = self
+            .client
+            .put(url.clone())
+            // Publish combines HTTP latency with the relay-side DHT PUT query.
+            .timeout(self.timeout * 3)
```
SeverinAlexB: Magic number? https://refactoring.guru/replace-magic-number-with-symbolic-constant

Also, why should this be 3 times higher? Should this be configurable independently?

pkarr/src/client/relays.rs
```diff
@@ -51,14 +43,16 @@ impl Debug for RelaysClient {
 impl RelaysClient {
     pub fn new(relays: Box<[Url]>, timeout: Duration) -> Self {
         let inflight_publish = InflightPublishRequests::new(relays.len());
+        let relays = relays
+            .into_vec()
+            .into_iter()
+            .map(|url| {
+                RelayClient::new(url, timeout).expect("Client building should be infallible")
```
SeverinAlexB: Are you sure this is always infallible? What if the reqwest client throws an error, for example can't allocate a socket?

pkarr/src/relay_client.rs
```diff
@@ -0,0 +1,647 @@
...
+        request.send().await.map_err(RelayError::from_reqwest)
+    }
+
+    fn build_url(&self, public_key: &PublicKey, policy: Option<ResolvePolicy>) -> Url {
+        let mut url = self.base_url.clone();
+
+        {
+            let mut segments = url
+                .path_segments_mut()
+                .expect("relay URL cannot be a base URL");
```
SeverinAlexB, line 185: Not sure I am a fan of the expect here.

Also, probably smart to validate this in the `new()` method and return (not panic) the error early compared to doing this in the publish/resolve method where build_url is called.

pkarr/src/relay_client.rs
```diff
@@ -0,0 +1,647 @@
...
+            StatusCode::BAD_REQUEST => Self::BadRequest,
+            StatusCode::CONFLICT => Self::NotMostRecent,
+            StatusCode::PRECONDITION_FAILED => Self::CasFailed,
+            StatusCode::PRECONDITION_REQUIRED => Self::ConflictRisk,
+            status => Self::UnexpectedStatus(status),
+        }
+    }
+}
+
+fn memento_datetime(response: &Response) -> Option<Timestamp> {
```
SeverinAlexB: How about a more expressive name?

```suggestion
fn extract_memento_datetime(response: &Response) -> Option<Timestamp> {
```

SeverinAlexB (changes requested):
A few questions, otherwise it's looking good.

pkarr/src/relay_client.rs
andrei-21: For now I moved this magic to the pkarr client, I will address it properly when I refactor the pkarr client.

pkarr/src/client/relays.rs
andrei-21: Fixed.

pkarr/src/relay_client.rs
andrei-21: Added a check in `new()`.

pkarr/src/relay_client.rs
andrei-21: Done.

pkarr/src/client/relays.rs
```diff
@@ -49,18 +41,24 @@ impl Debug for RelaysClient {
...
 impl RelaysClient {
-    pub fn new(relays: Box<[Url]>, timeout: Duration) -> Self {
+    pub fn new(relays: Box<[Url]>, timeout: Duration) -> Result<Self, RelayError> {
         let inflight_publish = InflightPublishRequests::new(relays.len());
-
-        Self {
+        let resolve_timeout = timeout;
+        // Publish combines HTTP latency with the relay-side DHT PUT query.
+        let publish_timeout = resolve_timeout
+            .checked_mul(3)
```
SeverinAlexB, line 49: NIT: Can we add a `// TODO: Fix later` here so we don't forget?

SeverinAlexB (approved):
Let's go. One NIT if you want to please my autistic self but I am fine with merging it without.

PR #251, opened 2026-05-27
fix(relay): add rustls feature to reqwest

andrei-21:
Without it reqwest fails to recognise https schemes with:
`invalid URL, scheme is not http`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,6 @@ resolver = "3"
 name = "pkarr"
 
 [workspace.dependencies]
-reqwest = { version = "0.13", default-features = false }
+reqwest = { version = "0.13", default-features = false, features = ["rustls"] }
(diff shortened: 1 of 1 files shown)


SeverinAlexB (approved):

Cargo.toml
```diff
@@ -9,6 +9,6 @@ resolver = "3"
 name = "pkarr"
 
 [workspace.dependencies]
-reqwest = { version = "0.13", default-features = false }
+reqwest = { version = "0.13", default-features = false, features = ["rustls"] }
```
86667, line 12: Can we also leave a comment to explain why it is needed please

andrei-21: I will add a comment here and a test for that after https://github.com/pubky/pkarr/pull/250.

PR #254, opened 2026-05-28
feat(relay)!: split HTTP and DHT rate limiting

andrei-21:
Build the relay DHT client directly from mainline config and add
separate rate limits for HTTP requests, inbound DHT requests, and
user-triggered DHT operations.

BREAKING CHANGE: Existing relay config files with a [rate_limiter]
table must add the required dht_quota field. The relay now rejects
rate-limiter configs that omit dht_quota instead of applying a default
DHT peer request limit.

BREAKING CHANGE: RelayBuilder::pkarr has been removed. Use
RelayBuilder::dht and RelayBuilder::request_timeout to configure the
relay DHT client.

### DIFF
--- a/pkarr/src/client/tests.rs
+++ b/pkarr/src/client/tests.rs
@@ -1,6 +1,6 @@
 //! Client native tests
 
-use std::net::Ipv4Addr;
+use std::net::{Ipv4Addr, ToSocketAddrs};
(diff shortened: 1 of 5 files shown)


SeverinAlexB (approved):
LGTM

PR #259, opened 2026-06-12
feat(dht)!: add DHT query reports

andrei-21:
Add resolve diagnostics for mutable GET queries and return publish
storage counts from PUT queries so relays can report suspicious DHT
outcomes.

BREAKING CHANGE: DHT publish and resolve APIs now expose query report
data and return shapes changed accordingly.

### DIFF
--- a/pkarr/src/dht/client.rs
+++ b/pkarr/src/dht/client.rs
@@ -8,7 +8,10 @@ use ntimestamp::Timestamp;
 
 use crate::{PublicKey, SignedPacket};
 
-use super::{DhtInfo, PublishError, ResolveError, ResolveOutcome, ResolveReport, ResolveResponse};
+use super::{
+    DhtInfo, PublishError, ResolveError, ResolveOutcome, ResolveReport, ResolveResponse,
+    ResolveValue,
+};
 
 /// Pkarr DHT client.
 #[derive(Clone, Debug)]
@@ -58,11 +61,22 @@ impl DhtClient {
 
     /// Resolve signed packets newer than the given timestamp and return query diagnostics.
     ///
+    /// The first valid signed packet is available immediately from the returned
+    /// [`ResolveResponse`]. Completing the response returns the most recent DHT
+    /// mutable value observed, which may be a newer mutable item that does not
+    /// contain a valid signed packet.
+    ///
+    /// BEP44 sequence numbers define mutable-item freshness. If a mutable item
+    /// with a higher sequence number is not a valid signed packet, that item is
+    /// still treated as the current DHT state and supersedes older valid signed
+    /// packets for the same key.
+    ///
     /// # Errors
     ///
     /// Returns an error when no signed packet is found or the DHT query did not
     /// receive enough useful responses to distinguish a miss from a query
-    /// failure.
+    /// failure. If DHT nodes return mutable values for the key but none verify
+    /// as signed packets, returns the newest observed mutable item sequence number.
     pub async fn resolve(
         &self,
         public_key: &PublicKey,
@@ -73,24 +87,29 @@ impl DhtClient {
             self.inner
                 .get_mutable_detailed(public_key.as_bytes(), None, more_recent_than);
 
-        let mut invalid_signed_packet_count = 0;
+        let mut highest_seq = None;
         while let Some(item) = detailed.items.next().await {
-            match SignedPacket::try_from(item) {
-                Ok(first) => {
-                    let most_recent = first.clone();
+            let seq = item.seq();
+            // BEP44 sequence numbers define the mutable item's freshness.
+            // A newer item that is not a valid signed packet still supersedes
+            // older valid signed packets for this key.
+            if seq >= highest_seq.unwrap_or(seq) {
+                highest_seq = Some(seq);
+                if let Ok(packet) = SignedPacket::try_from(&item) {
+                    let most_recent = packet.clone();
                     return Ok(ResolveResponse::new(
-                        first,
-                        finish_resolve(detailed, most_recent, invalid_signed_packet_count),
+                        packet,
+                        finish_resolve(detailed, most_recent),
                     ));
                 }
-                Err(_) => invalid_signed_packet_count += 1,
             }
         }
 
-        let report = ResolveReport::with_invalid_signed_packets(
-            detailed.outcome.recv().await,
-            invalid_signed_packet_count,
-        );
+        if let Some(seq) = highest_seq {
+            return Err(ResolveError::InvalidSignedPacket { seq });
+        }
+
+        let report = ResolveReport::new(detailed.outcome.recv().await);
         Err(report.into())
     }
 
@@ -151,23 +170,38 @@ impl TryFrom<MutableItem> for SignedPacket {
 async fn finish_resolve(
     mut detailed: GetMutableDetailed,
     mut most_recent: SignedPacket,
-    mut invalid_signed_packet_count: u32,
 ) -> ResolveOutcome {
+    let mut highest_seq = most_recent.timestamp().as_u64() as i64;
+
     while let Some(item) = detailed.items.next().await {
-        match SignedPacket::try_from(item) {
-            Ok(packet) if packet.more_recent_than(&most_recent) => most_recent = packet,
-            Ok(_) => {}
-            Err(_) => invalid_signed_packet_count += 1,
+        let seq = item.seq();
+        if seq >= highest_seq {
+            highest_seq = seq;
+            match SignedPacket::try_from(&item) {
+                Ok(packet) if packet.more_recent_than(&most_recent) => most_recent = packet,
+                _ => (),
+            }
         }
     }
 
-    let report = ResolveReport::with_invalid_signed_packets(
-        detailed.outcome.recv().await,
-        invalid_signed_packet_count,
-    );
+    let most_recent = completed_resolve_value(most_recent, highest_seq);
+    let report = ResolveReport::new(detailed.outcome.recv().await);
+
     ResolveOutcome {
         most_recent,
         report,
+    }
+}
+
+fn completed_resolve_value(most_recent: SignedPacket, highest_seq: i64) -> ResolveValue {
+    let packet_seq = most_recent.timestamp().as_u64() as i64;
+
+    if packet_seq >= highest_seq {
+        ResolveValue::ValidSignedPacket {
+            packet: most_recent,
+        }
+    } else {
+        ResolveValue::InvalidSignedPacket { seq: highest_seq }
     }
 }
 
@@ -201,5 +235,43 @@ mod tests {
         );
 
         assert_eq!(item, expected);
+    }
+
+    #[test]
+    fn completed_resolve_value_preserves_valid_signed_packet_when_invalid_seq_is_not_newer() {
+        let signed_packet = SignedPacket::builder()
+            .timestamp(10.into())
+            .address(
+                "_derp_region.iroh.".try_into().unwrap(),
+                "1.1.1.1".parse().unwrap(),
+                30,
+            )
+            .sign(&Keypair::random())
+            .unwrap();
+
+        assert_eq!(
+            super::completed_resolve_value(signed_packet.clone(), 10),
+            super::ResolveValue::ValidSignedPacket {
+                packet: signed_packet
+            }
+        );
+    }
+
+    #[test]
+    fn completed_resolve_value_respects_newer_invalid_signed_packet_seq() {
+        let signed_packet = SignedPacket::builder()
+            .timestamp(10.into())
+            .address(
+                "_derp_region.iroh.".try_into().unwrap(),
+                "1.1.1.1".parse().unwrap(),
+                30,
+            )
+            .sign(&Keypair::random())
+            .unwrap();
+
+        assert_eq!(
+            super::completed_resolve_value(signed_packet, 11),
+            super::ResolveValue::InvalidSignedPacket { seq: 11 }
+        );
     }
 }
--- a/pkarr/src/dht/client.rs
+++ b/pkarr/src/dht/client.rs
@@ -1,11 +1,19 @@
+use std::future::Future;
+
 use ed25519_dalek::Signature;
 use futures_lite::{Stream, StreamExt};
-use mainline::{async_dht::AsyncDht, Config, Dht, MutableItem};
+use mainline::{
+    async_dht::{AsyncDht, GetMutableDetailed},
+    Config, Dht, MutableItem,
+};
 use ntimestamp::Timestamp;
 
 use crate::{PublicKey, SignedPacket};
 
-use super::{DhtInfo, PublishError};
+use super::{DhtInfo, PublishError, ResolveFound, ResolveReport};
+
+/// Minimum DHT nodes expected to acknowledge storing a published packet.
+pub const MINIMUM_PUBLISH_STORED_NODES: u32 = 10;
 
 /// Pkarr DHT client.
 #[derive(Clone, Debug)]
@@ -27,14 +35,17 @@ impl DhtClient {
         &self,
         signed_packet: &SignedPacket,
         cas: Option<Timestamp>,
-    ) -> Result<(), PublishError> {
+    ) -> Result<u32, PublishError> {
         let cas = cas.map(|timestamp| timestamp.as_u64() as i64);
-        self.inner.put_mutable(signed_packet.into(), cas).await?;
-        Ok(())
+        Ok(self
+            .inner
+            .put_mutable(signed_packet.into(), cas)
+            .await?
+            .stored_at)
     }
 
     /// Resolve signed packets newer than the given timestamp.
-    pub fn resolve(
+    pub fn resolve_stream(
         &self,
         public_key: &PublicKey,
         more_recent_than: Option<Timestamp>,
@@ -44,6 +55,45 @@ impl DhtClient {
         self.inner
             .get_mutable(public_key.as_bytes(), None, more_recent_than)
             .filter_map(|item| SignedPacket::try_from(item).ok())
+    }
+
+    /// Resolve signed packets newer than the given timestamp and return query diagnostics.
+    pub async fn resolve(
+        &self,
+        public_key: &PublicKey,
+        more_recent_than: Option<Timestamp>,
+    ) -> Result<
+        ResolveFound<impl Future<Output = (SignedPacket, ResolveReport)> + Send>,
+        ResolveReport,
+    > {
+        let more_recent_than = more_recent_than.map(|timestamp| timestamp.as_u64() as i64);
+        let mut detailed =
+            self.inner
+                .get_mutable_detailed(public_key.as_bytes(), None, more_recent_than);
+
+        let mut invalid_signed_packet_count = 0;
+        while let Some(item) = detailed.items.next().await {
+            match SignedPacket::try_from(item) {
+                Ok(first) => {
+                    let most_recent = first.clone();
+                    return Ok(ResolveFound {
+                        first,
+                        completion: finish_resolve(
+                            detailed,
+                            most_recent,
+                            invalid_signed_packet_count,
+                        ),
+                    });
+                }
+                Err(_) => invalid_signed_packet_count += 1,
+            }
+        }
+
+        let report = ResolveReport::with_invalid_signed_packets(
+            detailed.outcome.recv().await,
+            invalid_signed_packet_count,
+        );
+        Err(report)
     }
 
     /// Return information about the underlying DHT node.
@@ -98,6 +148,27 @@ impl TryFrom<MutableItem> for SignedPacket {
     fn try_from(i: MutableItem) -> Result<Self, crate::errors::SignedPacketVerifyError> {
         SignedPacket::try_from(&i)
     }
+}
+
+async fn finish_resolve(
+    mut detailed: GetMutableDetailed,
+    mut most_recent: SignedPacket,
+    mut invalid_signed_packet_count: u32,
+) -> (SignedPacket, ResolveReport) {
+    while let Some(item) = detailed.items.next().await {
+        match SignedPacket::try_from(item) {
+            Ok(packet) if packet.more_recent_than(&most_recent) => most_recent = packet,
+            Ok(_) => {}
+            Err(_) => invalid_signed_packet_count += 1,
+        }
+    }
+
+    let report = ResolveReport::with_invalid_signed_packets(
+        detailed.outcome.recv().await,
+        invalid_signed_packet_count,
+    );
+
+    (most_recent, report)
 }
 
 #[cfg(test)]
--- a/pkarr/src/dht/client.rs
+++ b/pkarr/src/dht/client.rs
@@ -1,5 +1,3 @@
-use std::future::Future;
-
 use ed25519_dalek::Signature;
 use futures_lite::{Stream, StreamExt};
 use mainline::{
@@ -10,7 +8,7 @@ use ntimestamp::Timestamp;
 
 use crate::{PublicKey, SignedPacket};
 
-use super::{DhtInfo, PublishError, ResolveFound, ResolveReport};
+use super::{DhtInfo, PublishError, ResolveError, ResolveOutcome, ResolveReport, ResolveResponse};
 
 /// Minimum DHT nodes expected to acknowledge storing a published packet.
 pub const MINIMUM_PUBLISH_STORED_NODES: u32 = 10;
@@ -31,6 +29,10 @@ impl DhtClient {
     }
 
     /// Publish a signed packet to the DHT.
+    ///
+    /// # Returns
+    ///
+    /// The number of DHT nodes that acknowledged storing the packet.
     pub async fn publish(
         &self,
         signed_packet: &SignedPacket,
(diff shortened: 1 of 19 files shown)


pkarr/src/dht/client.rs
```diff
@@ -1,11 +1,19 @@
...
+};
 use ntimestamp::Timestamp;
 
 use crate::{PublicKey, SignedPacket};
 
-use super::{DhtInfo, PublishError};
+use super::{DhtInfo, PublishError, ResolveFound, ResolveReport};
+
+/// Minimum DHT nodes expected to acknowledge storing a published packet.
+pub const MINIMUM_PUBLISH_STORED_NODES: u32 = 10;
```
SeverinAlexB: This will not work with testnet/private DHTs where you potentially have way fewer nodes like 3 or even just 1 node? So far, we avoided any fixed number except for `MAX_BUCKET_SIZE_K` in the mainline crate due to this. `DEFAULT_MINIMUM_QUERIED_NODES` seems to be a duplication of `MAX_BUCKET_SIZE_K`. Maybe we need to export `MAX_BUCKET_SIZE_K` from mainline?

pkarr/src/dht/client.rs
```diff
@@ -27,14 +35,17 @@ impl DhtClient {
         &self,
         signed_packet: &SignedPacket,
         cas: Option<Timestamp>,
-    ) -> Result<(), PublishError> {
+    ) -> Result<u32, PublishError> {
```
SeverinAlexB, line 37: This method lacks a doc string describing what the return value means.

pkarr/src/dht/client.rs
```diff
@@ -46,6 +57,45 @@ impl DhtClient {
...
 
+    /// Resolve signed packets newer than the given timestamp and return query diagnostics.
+    pub async fn resolve(
+        &self,
+        public_key: &PublicKey,
+        more_recent_than: Option<Timestamp>,
+    ) -> Result<
+        ResolveFound<impl Future<Output = (SignedPacket, ResolveReport)> + Send>,
+        ResolveReport,
+    > {
```
SeverinAlexB: `ResolveFound<impl Future<Output = (SignedPacket, ResolveReport)> + Send>` is a very noisy public API. Can we simplify this? DhtClient is part of the public API so this matters.

pkarr/src/dht/client.rs
```diff
@@ -46,6 +57,45 @@ impl DhtClient {
...
     }
 
+    /// Resolve signed packets newer than the given timestamp and return query diagnostics.
+    pub async fn resolve(
+        &self,
+        public_key: &PublicKey,
+        more_recent_than: Option<Timestamp>,
+    ) -> Result<
+        ResolveFound<impl Future<Output = (SignedPacket, ResolveReport)> + Send>,
+        ResolveReport,
```
SeverinAlexB: The `ResolveReport` is an error? At the same time `ResolveFound` also returns a report? Without thinking about this deeply, this sounds conceptually incorrect.

Also, it's very hard to interpret ResolveReport as an error. Right now, it feels like I need to study this object deeply to figure out if the packet was just not found, or invalid, or what really happened.

pkarr/src/dht/client.rs
```diff
@@ -46,6 +57,45 @@ impl DhtClient {
             .filter_map(|item| SignedPacket::try_from(item).ok())
     }
 
+    /// Resolve signed packets newer than the given timestamp and return query diagnostics.
+    pub async fn resolve(
+        &self,
+        public_key: &PublicKey,
+        more_recent_than: Option<Timestamp>,
+    ) -> Result<
+        ResolveFound<impl Future<Output = (SignedPacket, ResolveReport)> + Send>,
```
SeverinAlexB: This might be a matter of taste: Should a "Not Found" really be an error?

pkarr/src/dht/resolve.rs
```diff
@@ -0,0 +1,383 @@
+use crate::SignedPacket;
+
+/// Successful immediate DHT resolve result.
+pub struct ResolveFound<F> {
+    /// First valid signed packet returned by the DHT query.
+    pub first: SignedPacket,
+    /// Future that drains the rest of the query and returns the most recent packet and report.
+    pub completion: F,
+}
```
SeverinAlexB: This is an interesting pattern: Provide first packet directly with the option for completion 👍

pkarr/src/dht/resolve_report.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+/// Default minimum number of DHT nodes a resolve query should visit.
+const DEFAULT_MINIMUM_QUERIED_NODES: u32 = 20;
+
+/// Default minimum number of DHT nodes that should respond to a resolve query.
+const DEFAULT_MINIMUM_RESPONDED_NODES: u32 = 3;
+
+/// Default minimum number of valid responses expected from a resolve query.
+const DEFAULT_MINIMUM_VALID_RESPONSES: u32 = 1;
```
SeverinAlexB: `DEFAULT_MINIMUM_QUERIED_NODES` feels like a duplicate of the bucket size from mainline.

`DEFAULT_MINIMUM_RESPONDED_NODES` Does not work in DHT network with fewer nodes, for example in private or testnet DHTs.

pkarr/src/dht/resolve.rs
```diff
@@ -0,0 +1,383 @@
...
+            outcome,
+            invalid_signed_packet_count,
+        }
+    }
+
+    fn suspicions(
+        &self,
+        outcome: &mainline::GetMutableOutcome,
+        invalid_signed_packet_count: u32,
+    ) -> Vec<ResolveSuspicion> {
```
SeverinAlexB: What is the definition of suspicious? This word feels weak descriptively. This is a question I have with all uses of suspicious in this whole PR.

pkarr/src/relay_client.rs
```diff
@@ -241,6 +241,18 @@ pub enum RelayError {
...
+    NoValidDhtResponses,
+
+    /// Relay did not query any DHT nodes.
+    #[error("relay queried no DHT nodes")]
+    NoDhtNodesQueried,
+
+    /// Relay queried DHT nodes, but none responded.
+    #[error("relay DHT query received no responses")]
+    NoDhtResponses,
+
```
SeverinAlexB: What is `NoValidDhtResponses`? How does it differ from `InvalidSignedPacket`? Is `InvalidSignedPacket` still necessary?

relay/src/error.rs
```diff
@@ -74,13 +74,31 @@ impl From<dht::PublishError> for Error {
...
+        } else if report.responded() == 0 {
+            Error::new(
+                StatusCode::GATEWAY_TIMEOUT,
+                "No queried DHT nodes responded",
+            )
+        } else if report.valid_responses() == 0 {
+            Error::new(
+                StatusCode::BAD_GATEWAY,
+                "No responded DHT nodes returned valid response",
+            )
```
SeverinAlexB: You're adding gateway errors here. Are you sure they will play nicely in case the relay is actually behind a gateway (reverse proxy, load balancer) and these errors will not be misinterpreted?

SeverinAlexB (changes requested):
I need to do a second review after we completed these comments. My main points are: The code is pretty unclean/hard to interpret + hardcoded constants will not work with smaller DHT networks.

pkarr/src/relay_client.rs
andrei-21: `NoValidDhtResponses` includes `InvalidSignedPacket`. `InvalidSignedPacket` is not necessary here, that is why it is not here.
The caveat is that mainline dht client returns `NoValidDhtResponses`, but when we parse the items we can get an invalid signed packet (still a valid mutable item from mainline point of view), that is why we have to carefully add this instance to `NoValidDhtResponses`.

pkarr/src/dht/client.rs
andrei-21: Fixed.

pkarr/src/dht/client.rs
andrei-21: Right, it is confusing. I added `ResolveError` which is build from `ResolveReport` when needed.

relay/src/error.rs
andrei-21: Right, I replace them with `503 Service Unavailable`. We will lose the information at the status-code level, but the response body explains what actually happened.

pkarr/src/dht/client.rs
andrei-21: Fair, I implemented `ResolveResponse` to catch the first packet and competition. Now the return type is `Result<ResolveResponse, ResolveError>`.

pkarr/src/dht/client.rs
andrei-21: It is a matter of taste, see how nicely it plays out in `handler.rs`.

pkarr/src/dht/client.rs
andrei-21: `MAX_BUCKET_SIZE_K` is not exposed from `mainline`. So 20 here is an educated guess about how a healthy network behaves.
Regarding testnet, I agree. I exposed `ResolveReportPolicy` with the `testnet()` constructor so the client can decide what policy is appropriate for them.

pkarr/src/dht/resolve_report.rs
andrei-21: Repeating:
`MAX_BUCKET_SIZE_K` is not exposed from mainline. So 20 here is an educated guess about how a healthy network behaves.
Regarding testnet, I agree. I exposed `ResolveReportPolicy` with the `testnet()` constructor so the client can decide what policy is appropriate for them.

pkarr/src/dht/resolve.rs
andrei-21: I renamed them to **warnings**.

pkarr/src/dht/client.rs
SeverinAlexB: > So 20 here is an educated guess about how a healthy network behaves.

Please read yourself into the mainline BEPs. The bucket size is a very specific number, not an educated guess (k=8 in the BEPs, k=20 in our mainline lib). Make sure to also research how our mainline lib decides how many nodes it will publish the packet to. It may be more or less than 20 depending on the DHT condition.

> MAX_BUCKET_SIZE_K is not exposed from mainline

See my initial comment: "Maybe we need to export MAX_BUCKET_SIZE_K from mainline?"

> Regarding testnet, I agree. I exposed ResolveReportPolicy with the testnet() constructor so the client can decide what policy is appropriate for them.

Again, so far we specifically avoided hard coded numbers. Your `testnet()` constructor does not cover private DHT networks that can have any size. I think our mainline lib defines insufficient publishing as  < K/2.

So my question is: What is the advantage of having hard coded numbers here? Does the upside of having a hard coded number outweight the downside of complicating the constructor? See my new comment in https://github.com/pubky/pkarr/pull/259#discussion_r3428001706 as a continuation.

pkarr/src/dht/resolve_report.rs
SeverinAlexB: A DHT network can be anywhere from 1 node to 20 million nodes+. It's very hard to make hardcoded values work. The testnet constant `TESTNET_MINIMUM_QUERIED_NODES` below already fails because a lot of testnets are only one node. At the same time, you need to think about how to make the API intuitive so the dev using pkarr doesn't require a lot of cognitive load.

I think one thing that you might confuse here: Mainnet and Testnet in mainline is not the same as in Bitcoin. A mainline Testnet is like a Bitcoin Regtest. It is not shared with others like the Bitcoin Testnet. A mainline testnet therefore has no standardized size. In the pubky-core tests, every test spins up a new mainline Testnet with 1 node.

With the pkarr client being the most used struct, how would the constructor/builder look like that configures this for mainnet/private and testnet DHT with a variable DHT size? What I am thinking is: Mainnet hard coded values derived from the MAX_BUCKET_SIZE_K might be worth it. Private/Testnet values are too hard to nail done as constants though. So if we can find a clean way in the constructor/builder to set this, this could work. Maybe we can even use the DHT diagnostic/statistics to remove the configuration need? Lots of room to explore

pkarr/src/dht/error.rs
```diff
@@ -54,3 +56,85 @@ impl From<PutMutableError> for PublishError {
...
+    NoNodesResponded,
+
+    /// DHT query did not receive any valid responses.
+    #[error("no responded DHT nodes returned valid response")]
+    NoValidResponses,
+
+    /// No signed packet was found.
+    #[error("signed packet not found")]
+    NotFound,
+}
```
SeverinAlexB, line 85: Should the ResolveReport be part of the error so users can inspect the problem deeper if they want to?

```rust
pub enum ResolveError {
    #[error("no DHT nodes were queried")]
    NoNodesQueried { report: ResolveReport },

    #[error("no queried DHT nodes responded")]
    NoNodesResponded { report: ResolveReport },

    #[error("no responded DHT nodes returned valid response")]
    NoValidResponses { report: ResolveReport },

    #[error("signed packet not found")]
    NotFound { report: ResolveReport },
}

impl ResolveError {
    pub fn report(&self) -> &ResolveReport {
        match self {
            Self::NoNodesQueried { report }
            | Self::NoNodesResponded { report }
            | Self::NoValidResponses { report }
            | Self::NotFound { report } => report,
        }
    }
}
```

pkarr/src/dht/resolve_report.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// Minimum number of unique DHT nodes that should be queried.
+    pub minimum_queried_nodes: u32,
+    /// Minimum number of queried DHT nodes that should respond.
+    pub minimum_responded_nodes: u32,
+    /// Minimum number of responses with a valid mutable GET shape.
+    pub minimum_valid_responses: u32,
+    /// Whether invalid values or invalid response shapes should be reported as warnings.
+    pub warn_on_invalid_responses: bool,
+    /// Whether KRPC error responses should be reported as warnings.
+    pub warn_on_krpc_errors: bool,
```
SeverinAlexB: I do like the term "warning" more than "suspicion" but "warn_on" sounds like it will log the problem. Can we get a better name for this? Maybe `report_invalid_responses` or `store_invalid_responses`?

On another note: Something feels off about the way this code is architectured but I have a hard time nailing it down.

Maybe it's this: When I look at `ResolveWarning`, some of them don't seem actionable for a dev using pkarr as a lib. The warnings are technically correct but hard to interpret.

- `TooFewNodesQueried` If the user get's this error, what should they do? Is the DHT connection faulty? Is the configured parameter wrong?
- `TooFewNodesResponded` same here, what is the actionable advice when the user gets this?
- `TooFewValidResponses` What does too few mean? In theory, one valid response is enough (constant is set to 1 anyway). Is there ever a case where I would require 2+ valid responses?
- `InvalidResponses` This feels actionable: Somebody published things that are invalid. But, from the variable names and doc strings, I don't see the difference between `invalid_values` and `invalid_responses`. Also, is one of them a protocol error or are both pkarr related errors?
- `KrpcErrors` This makes sense. It is actionable is a sense that the nodes returned a protocol error. Nothing the user can do about. Either our code is faulty or the peers code.

relay/src/handlers.rs
```diff
@@ -22,21 +23,24 @@ pub async fn put(
     body: Bytes,
 ) -> Result<impl IntoResponse, Error> {
     let signed_packet = pkarr::SignedPacket::from_relay_payload(&public_key, &body)
```
SeverinAlexB, line 25: I just saw the `from_relay_payload` method for the first time. Is this a public method in every pkarr client?

Looks like Ar added this 2 years ago. Relay concerns shouldn't be part of the SignedPacket. Not needed now but probably worth fixing in the future.

relay/src/handlers.rs
```diff
@@ -22,21 +23,24 @@ pub async fn put(
...
         }
     }
 
     enforce_user_dht_rate_limit(&state, real_ip.as_ref())?;
 
-    state.dht.publish(&signed_packet, cas).await?;
+    let stored_at = state.dht.publish(&signed_packet, cas).await?;
+    if stored_at < MINIMUM_PUBLISH_STORED_NODES {
+        warn!(public_key = ?signed_packet.public_key(), stored_at, "DHT publish completed with warnings");
+    }
```
SeverinAlexB: This will always throw a warning if the relay is run with a testnet like we do with pubky-core testnet.

relay/src/handlers.rs
```diff
@@ -222,6 +228,17 @@ fn decrement(timestamp: Timestamp) -> Option<Timestamp> {
...
+fn log_resolve_warnings(public_key: &pkarr::PublicKey, report: &ResolveReport) {
+    let warnings = report.classify_warnings(ResolveReportPolicy::default());
+    if !warnings.is_empty() {
+        warn!(
+            ?public_key,
+            ?warnings,
+            "DHT resolve completed with warnings"
+        );
+    }
+}
```
SeverinAlexB: The relay can be used in any environment (mainnet, testnet aso). So just using the default ResolveReportPolicy does not work.

pkarr/src/relay_client.rs
SeverinAlexB: > InvalidSignedPacket is not necessary here, that is why it is not here.

`InvalidSignedPacket` is in `RelayError`. See line 222.

SeverinAlexB (changes requested):

pkarr/src/relay_client.rs
andrei-21: Oh, that is completely different:
```rust
    /// Relay returned a malformed signed packet payload.
    #[error("relay returned an invalid signed packet: {0}")]
    InvalidSignedPacket(crate::errors::SignedPacketVerifyError),
```
It has nothing to do with DHT directly. It means that the relay is misbehaving and returned an invalid packet, but it had to validate itself.

pkarr/src/dht/report_policy.rs
```diff
@@ -0,0 +1,349 @@
...
+const TESTNET_MINIMUM_PUBLISH_STORED_NODES: u32 = 1;
+
+/// Testnet minimum number of DHT nodes a resolve query should visit.
+const TESTNET_MINIMUM_QUERIED_NODES: u32 = 1;
+
+/// Testnet minimum number of DHT nodes that should respond to a resolve query.
+const TESTNET_MINIMUM_RESPONDED_NODES: u32 = 1;
+
+/// Testnet minimum number of usable responses expected from a resolve query.
+const TESTNET_MINIMUM_USABLE_RESPONSES: u32 = 1;
```
SeverinAlexB: NIT: Can we get a clearer separation of these consts? The doc strings of the consts feel like a duplicate of the ReportPolicy variable docstrings. At the same time, it's probably worth grouping the mainnet/testnet consts better.

One way but not the only way to group them would be:

```
impl ReportPolicy {
    const MAINNET: Self = Self {
        minimum_publish_stored_nodes: 10,
        minimum_queried_nodes: 20,
        minimum_responded_nodes: 5,
        minimum_usable_dht_responses: 3,
        report_malformed_responses: true,
        report_krpc_errors: true,
    };

    const TESTNET: Self = Self {
        minimum_publish_stored_nodes: 1,
        minimum_queried_nodes: 1,
        minimum_responded_nodes: 1,
        minimum_usable_dht_responses: 1,
        report_malformed_responses: true,
        report_krpc_errors: true,
    };

    pub fn mainnet() -> Self {
        Self::MAINNET
    }

    pub fn testnet() -> Self {
        Self::TESTNET
    }
}
```

pkarr/src/dht/report_policy.rs
```diff
@@ -0,0 +1,349 @@
...
+const MINIMUM_PUBLISH_STORED_NODES: u32 = 10;
+
+/// Default minimum number of DHT nodes a resolve query should visit.
+const DEFAULT_MINIMUM_QUERIED_NODES: u32 = 20;
+
+/// Default minimum number of DHT nodes that should respond to a resolve query.
+const DEFAULT_MINIMUM_RESPONDED_NODES: u32 = 5;
+
+/// Default minimum number of usable responses expected from a resolve query.
+const DEFAULT_MINIMUM_USABLE_RESPONSES: u32 = 3;
```
SeverinAlexB: Isn't `DEFAULT_MINIMUM_USABLE_RESPONSES = 1` good enough here? DHT packets decay naturally. One valid mutable get is as good as 3, no? If it is `1` then this is equivalent to "Not found" anyway, no?

pkarr/src/dht/resolve_response.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// The most recent mutable item is newer than all valid signed packets, but
+    /// does not contain a valid signed packet.
+    ///
+    /// Older valid signed packets must not be treated as current when this
+    /// variant is returned; the newer invalid mutable item is the current DHT
+    /// state for the key.
+    InvalidSignedPacket {
+        /// Mutable item sequence number.
+        seq: i64,
+    },
```
SeverinAlexB: This is a very weird state. It feels like the end user should not worry about this. This case should be handled internally to make the public API simpler.

`ResolveValue.InvalidSignedPacket` is basically an error wrapped as a result?

relay/src/handlers.rs
```diff
@@ -59,19 +68,22 @@ pub async fn get(
...
         ResolvePolicy::CacheFirst => match cached_packet {
-            Some(packet) if !packet.is_expired(state.minimum_ttl, state.maximum_ttl) => {
-                Some(packet)
-            }
+            Some(packet) if !packet.is_expired(state.minimum_ttl, state.maximum_ttl) => packet,
             _ => {
                 enforce_user_dht_rate_limit(&state, real_ip.as_ref())?;
-                let mut stream = state.dht.resolve(&public_key, more_recent_than);
-                let packet = stream.next().await;
+                let result = state.dht.resolve(&public_key, more_recent_than).await?;
```
SeverinAlexB: This is not a Result but the ResolveResponse and should therefore be named `response`.

SeverinAlexB (changes requested):

relay/src/handlers.rs
andrei-21: Done.

pkarr/src/dht/report_policy.rs
andrei-21: Good idea. Done.

pkarr/src/dht/report_policy.rs
andrei-21: Usable here means that a node responded with a value, "no value," or "no more recent." If we have good connectivity to the network, several nodes should respond with a usable response even if it is "no value."
Now these numbers are mostly God's feelings; I think we will get better insights when we address #261.

pkarr/src/dht/resolve_response.rs
andrei-21: We cannot return it as an error here because we already could return a `ResolveReport` with a first signed packet (not the latest tho).
The user needs this information to invalidate their cache (I will implement it in following PRs).

relay/src/lib.rs
```diff
@@ -142,12 +147,13 @@ impl Relay {
     async unsafe fn run(config: RelayConfig) -> anyhow::Result<Self> {
         let dht_config = mainline_config(&config);
 
-        unsafe { Self::run_with_dht(config, dht_config) }.await
+        unsafe { Self::run_with_dht(config, dht_config, ReportPolicy::mainnet()) }.await
     }
```
SeverinAlexB, line 160: Should the policy be configurable in the `config.toml`? I think it should. Otherwise there is no way to configure a custom policy for the relay, not even programatically as far as I can see.

pkarr/src/dht/resolve_response.rs
SeverinAlexB: Not sure I understand your point. Why can't we do this?

```rust
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct ResolveOutcome {
    /// Most recent mutable item returned by the DHT query.
    pub most_recent: Result<SignedPacket, ResolveError>,
    /// Diagnostics collected from the DHT query.
    pub report: ResolveReport,
}
```

SeverinAlexB (changes requested):
Please also check the previous comments. Some of them have not been addressed yet.

Question: Should ResolvePolicy/ResolveReport be part of mainline and not pkarr? I am asking this question because it seems to measure regular DHT health and not really pkarr health.

pkarr/src/dht/error.rs
andrei-21: Not needed anymore. Now `resolve()` returns `ResolveResponse` (not a `Result<>`), it means that `ResolveOutcome::report` is always available.

pkarr/src/dht/resolve_response.rs
andrei-21: Done.

relay/src/handlers.rs
andrei-21: Yes, it is public https://docs.rs/pkarr/latest/pkarr/struct.SignedPacket.html#method.from_relay_payload.

relay/src/lib.rs
andrei-21: I see no real need for that for now, the toml config does not really allow to configure DHT node. In this sense pkarr relay is opinionated, your are suppose to run it in a good environment, otherwise there is no point to run it.

andrei-21:
> Should ResolvePolicy/ResolveReport be part of mainline and not pkarr? I am asking this question because it seems to measure regular DHT health and not really pkarr health.

It is not very obvious. But I view mainline as a general library which can be used in different environments and it should not make assumptions about that. So in `get_mitable()` it returns what it observed and you as a client decides if it is ok, suspicions or weird (base on the knowledge about the environments and your expectations).

relay/src/lib.rs
SeverinAlexB: Ok, the config.toml is hardcoded to mainnet at the moment (probably should allow to set bootstrap nodes in the future but no need at the moment).

How does a pubky-core testnet configure the relay so it uses testnet or a custom relay policy then?

andrei-21: I pushed a setter for `RelayBuilder`.

SeverinAlexB (approved):

PR #260, opened 2026-06-22
fix(relay): handle container shutdown signals

andrei-21:
Listen for SIGTERM in addition to Ctrl-C so Docker and Compose
shutdown requests stop the relay promptly.

Remove the Compose restart policy for local relay runs.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,6 @@
 services:
   pkarr-relay:
     build: .
-    restart: unless-stopped
     
     # need both TCP and UDP
     ports:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,7 +1,6 @@
 services:
   pkarr-relay:
     build: .
-    restart: unless-stopped
```
MCarlomagno, line 4: not sure if this is part of the scope of the PR, just marking in case it's a leftover

MCarlomagno (approved):

docker-compose.yml
andrei-21: It is just a clean up. The compose file is not for production, so we do not need it.

PR #262, opened 2026-06-23
fix(relays): use webpki-roots TLS for the relay HTTP client (Android verifier panic + revocation false-positives)

coreyphillips:
## Summary

The per-relay HTTP client is built with reqwest's default rustls config, which is `rustls-platform-verifier` (`pkarr/src/relay_client.rs` — `RelayClient::new` previously did `Client::builder().build()`). On Android this causes two failures, and **neither is fixable by the consumer**, because the relay client accepts no TLS config:

1. **Panic without init.** `rustls-platform-verifier` must be handed the `JavaVM` + app `Context` before the first handshake, or it panics with `Expect rustls-platform-verifier to be initialized`. Consumers loading pkarr via uniffi/JNA (where `JNI_OnLoad` never fires) have to add a JNI shim purely to initialize this.
2. **Revocation false-positives.** The platform verifier hard-fails revocation on Android. With Let's Encrypt's newer sharded-CRL hierarchy, valid relay certs are falsely rejected on some devices with `invalid peer certificate: Revoked`.

Relays are enabled by default (`ClientBuilder::default()` → `DEFAULT_RELAYS`) and are the primary resolution path on mobile, where mainline DHT (UDP) is often unavailable — so this is on the hot path for many apps.

## Fix

Preconfigure the relay client (native only) with an explicit webpki/Mozilla-roots rustls config using the **ring** provider (matching the rest of pkarr's TLS code) and **no** revocation checking — matching browsers' soft-fail behavior and requiring no platform-specific initialization.

- `pkarr/src/relay_client.rs` — `build_relay_http_client()`, gated on `cfg(not(wasm_browser))`. WASM keeps the plain builder (the browser performs TLS).
- `pkarr/Cargo.toml` — adds a native-only optional `webpki-roots = "1"` dep and wires `dep:rustls`/`dep:webpki-roots` into the `relays` feature. Both are declared only under the `cfg(not(target_family = "wasm"))` dependency table, so they are ignored on wasm.

This also means apps no longer need the platform verifier (and its Android init dance) just to use relays.

## Testing

Verified to compile across the relevant matrix:

- `cargo check -p pkarr` (native, default `full-client`)
- `cargo check -p pkarr --no-default-features --features relays` (native, minimal)
- `cargo check -p pkarr --target wasm32-unknown-unknown --no-default-features --features relays` (wasm — confirms the native-only deps are ignored and no rustls is pulled)
- `cargo test -p pkarr --no-default-features --features relays --lib --no-run`

## Notes

- Companion fix in pubky-core for the same root cause on its own ICANN client: pubky/pubky-core#456.
- Alternative designs if you'd prefer: expose a builder hook to inject a custom relay `reqwest::Client`/`rustls::ClientConfig`, rather than hard-coding webpki-roots. Happy to adjust.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1779,6 +1779,7 @@ dependencies = [
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "wasm-bindgen-test",
+ "webpki-roots",
 ]
 
 [[package]]
@@ -3217,6 +3218,15 @@ name = "webpki-root-certs"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f31141ce3fc3e300ae89b78c0dd67f9708061d1d2eda54b8209346fd6be9a92c"
+dependencies = [
+ "rustls-pki-types",
+]
+
+[[package]]
+name = "webpki-roots"
+version = "1.0.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf85cb06032201fa7c6f829d7db5a7e5aa45bcc0655327713065f6f0576731bf"
 dependencies = [
  "rustls-pki-types",
 ]
--- a/pkarr/Cargo.toml
+++ b/pkarr/Cargo.toml
@@ -82,6 +82,8 @@ page_size = { version = "0.6.0", optional = true }
 # feat: tls
 rustls = { workspace = true, optional = true }
 webpki = { package = "rustls-webpki", version = "0.103", optional = true }
+# feat: relays (native) — webpki/Mozilla root bundle for the relay HTTP client's TLS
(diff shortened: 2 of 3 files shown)


andrei-21:
Right now I am in the middle of a major refactoring. I see the issue you encountered; I will take it into account and provide a way to mitigate it.
I tend to think that `pkarr` is a generic library that should not mess with certificates, but it should let clients configure whichever `reqwest` client they want.

andrei-21:
Closing, as tweaking reqwest is not considered part of the library. The client can provide their `reqwest::Client` if they need customization: [`ClientBuilder::reqwest_client()`](https://github.com/pubky/pkarr/blob/main/pkarr/src/client/builder.rs#L306).

PR #264, opened 2026-07-02
fix(resolve)!: tighten relay and DHT state handling

andrei-21:
Return RelayError::NotFound for missing or unchanged relay resolves
while keeping higher-level relay streams treating missing relay responses
as absent packets.

Use cached valid packets as a floor for CacheFirst DHT results, but let
DhtNetworkOnly report the DHT state without cache suppression.

Use a shared reqwest client for relay HTTP requests and apply
DEFAULT_REQUEST_TIMEOUT to both relay and default DHT request configs.

BREAKING CHANGE: RelayClient::new now accepts a reqwest::Client, and
RelayClient::resolve returns Result<SignedPacket, RelayError> instead of
Result<Option<SignedPacket>, RelayError>.

### DIFF
--- a/pkarr/src/client/builder.rs
+++ b/pkarr/src/client/builder.rs
@@ -13,7 +13,7 @@ use crate::{errors::BuildError, Client};
 pub const DEFAULT_MAX_RECURSION_DEPTH: u8 = 7;
 
 /// Default request timeout for DHT and relay requests.
-pub const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
+pub const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(2);
 
 /// [Client]'s Config
 #[derive(Clone)]
@@ -60,7 +60,7 @@ impl Default for Config {
             cache: None,
 
             #[cfg(dht)]
-            dht: Some(default_dht_config(DEFAULT_REQUEST_TIMEOUT)),
+            dht: Some(make_dht_config(DEFAULT_REQUEST_TIMEOUT)),
 
             #[cfg(feature = "relays")]
             relays: Some(
@@ -142,7 +142,7 @@ impl ClientBuilder {
         F: FnOnce(&mut mainline::Config) -> &mut mainline::Config,
     {
         if self.0.dht.is_none() {
-            self.0.dht = Some(default_dht_config(self.0.request_timeout));
+            self.0.dht = Some(make_dht_config(self.0.request_timeout));
         }
 
         if let Some(ref mut builder) = self.0.dht {
@@ -295,7 +295,7 @@ impl ClientBuilder {
 }
 
 #[cfg(dht)]
-fn default_dht_config(request_timeout: Duration) -> mainline::Config {
+fn make_dht_config(request_timeout: Duration) -> mainline::Config {
     mainline::Config {
         request_timeout,
         ..Default::default()
--- a/pkarr/src/client/builder.rs
+++ b/pkarr/src/client/builder.rs
@@ -12,10 +12,8 @@ use crate::{errors::BuildError, Client};
 #[cfg(feature = "endpoints")]
 pub const DEFAULT_MAX_RECURSION_DEPTH: u8 = 7;
 
-#[cfg(dht)]
-pub const DEFAULT_REQUEST_TIMEOUT: Duration = mainline::DEFAULT_REQUEST_TIMEOUT;
-#[cfg(not(dht))]
-pub const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(2);
+/// Default request timeout for DHT and relay requests.
+pub const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
 
 /// [Client]'s Config
 #[derive(Clone)]
@@ -46,7 +44,7 @@ pub(crate) struct Config {
     ///
     /// The longer this timeout the longer resolve queries will take before consider failed.
     ///
-    /// Defaults to [DEFAULT_REQUEST_TIMEOUT]
+    /// Defaults to [DEFAULT_REQUEST_TIMEOUT].
     pub request_timeout: Duration,
 
     #[cfg(feature = "endpoints")]
@@ -62,7 +60,7 @@ impl Default for Config {
             cache: None,
 
             #[cfg(dht)]
-            dht: Some(mainline::Config::default()),
+            dht: Some(default_dht_config(DEFAULT_REQUEST_TIMEOUT)),
 
             #[cfg(feature = "relays")]
             relays: Some(
@@ -144,7 +142,7 @@ impl ClientBuilder {
         F: FnOnce(&mut mainline::Config) -> &mut mainline::Config,
     {
         if self.0.dht.is_none() {
-            self.0.dht = Some(Default::default());
+            self.0.dht = Some(default_dht_config(self.0.request_timeout));
         }
 
         if let Some(ref mut builder) = self.0.dht {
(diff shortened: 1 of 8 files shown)


pkarr/src/client/builder.rs
```diff
@@ -12,10 +12,8 @@ use crate::{errors::BuildError, Client};
 #[cfg(feature = "endpoints")]
 pub const DEFAULT_MAX_RECURSION_DEPTH: u8 = 7;
 
-#[cfg(dht)]
-pub const DEFAULT_REQUEST_TIMEOUT: Duration = mainline::DEFAULT_REQUEST_TIMEOUT;
-#[cfg(not(dht))]
-pub const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(2);
+/// Default request timeout for DHT and relay requests.
+pub const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
```
SeverinAlexB: Why would you increase the DHT timeout to 10s? That means, there is a very high change that every resolve_most_recent call will take 10s.

A most recent resolve makes calls to maybe 60 nodes. If only one of these nodes does not respond, you'll wait for 10s.

Maybe in the future, we need an adaptive timeout? https://chatgpt.com/share/6a465eec-59c0-83eb-9b43-d5a490133fb6
I feel like I have seen this before in the pkarr/mainline code. Not sure anymore though. Out of scope for this PR.

Let me know if I miss something

pkarr/src/client/builder.rs
```diff
@@ -296,6 +294,14 @@ impl ClientBuilder {
     }
 }
 
+#[cfg(dht)]
+fn default_dht_config(request_timeout: Duration) -> mainline::Config {
```
SeverinAlexB: I was confused when I read the `default_dht_config` method name? Can we get something more expressive here?

pkarr/src/types/resolve_policy.rs
```diff
@@ -21,6 +24,11 @@ pub enum ResolvePolicy {
     /// guarantee a fast response.
     /// May return an outdated packet in edge cases.
     ///
+    /// When a cached packet is expired and the DHT is queried, the cached
+    /// sequence is still used as a floor: older DHT packets are not returned,
+    /// and invalid DHT mutable items at or below the cached sequence are
+    /// treated as stale.
```
SeverinAlexB: Isn't this an implementation detail? It feels like the user doesn't care about this. He should just get the correct packet.

pkarr/src/types/resolve_policy.rs
```diff
@@ -29,6 +37,11 @@ pub enum ResolvePolicy {
     /// update the cache when that value contains a valid signed packet.
     /// This is slower, but more accurate.
     ///
+    /// This policy ignores cached packets while querying and interpreting the
+    /// DHT result. If the DHT currently contains an older valid packet than the
+    /// cache, that older packet is returned. If it contains an older invalid
+    /// mutable item, that invalid sequence is reported.
```
SeverinAlexB: Same here. Does this need to be in the doc string? Sounds like an unnecessary implementation detail that will only confuse users.

```suggestion
    /// cache, that older packet is returned.
```

relay/src/handlers.rs
```diff
@@ -61,23 +61,32 @@ pub async fn get(
     let key = CacheKey::from(&public_key);
 
     let policy = query.policy.unwrap_or(ResolvePolicy::CacheFirst);
-    let cached_packet = state.cache.get_read_only(&key);
-    let more_recent_than = cached_packet
+    let cached = state.cache.get(&key);
+    let more_recent_than = cached
         .as_ref()
         .map(SignedPacket::timestamp)
         .and_then(decrement);
```
SeverinAlexB: Why do we actually decrement the timestamp here?

relay/src/handlers.rs
```diff
@@ -61,23 +61,32 @@ pub async fn get(
...
+        (ResolvePolicy::CacheFirst, _) => {
+            enforce_user_dht_rate_limit(&state, real_ip.as_ref())?;
+            let response = state.dht.resolve(&public_key, more_recent_than).await;
+            let first_packet = response
+                .first()
+                .cloned()
+                .map(|packet| apply_cached_floor(Ok(packet), cached.as_ref()));
+
+            match first_packet {
+                Some(Ok(packet)) => {
```
SeverinAlexB: Having a match in match is a smell for a too deeply nested function. This whole `match (policy, cached.as_ref())` is very big. Worth refactoring IMO

SeverinAlexB (changes requested):
Looking good overall. I am not sure what timeout strategy is best here.

pkarr/src/client/builder.rs
andrei-21: True. I reverted it to 2 seconds to keep the status quo.

relay/src/handlers.rs
andrei-21: We want to get items with the same seq that we have in the cache to make sure that the item is still available in the DHT network and we can refresh `last_seen`.

pkarr/src/client/builder.rs
andrei-21: Renamed to `make_dht_config()`.

pkarr/src/types/resolve_policy.rs
andrei-21: That is the point of the PR. It is important because in a situation where we have a newer item in the cache than in the DHT network, we cannot simply return the _correct_ packet. We need to specify what exactly we do in such a situation.

pkarr/src/types/resolve_policy.rs
andrei-21: Applied.

relay/src/handlers.rs
andrei-21: Fair. I split it into multiple smaller functions.

pkarr/src/types/resolve_policy.rs
SeverinAlexB: That's what I mean. We need to decided here in the code and not overburden the user with doc string explanations. I do have a hard time understanding these details reading them here and I am a repo collaborator. How should a user understand this?

relay/src/handlers.rs
SeverinAlexB: Nice. 

NIT (no need to change if you don't want): The `handlers.rs` file has a lot of business logic in there aka a lot of functions that do not have anything to do with handlers (UI layer). They could be extracted in services to make it cleaner. For example:

CleanShot 2026-07-03 at 10 49 00@2x

pkarr/src/types/resolve_policy.rs
andrei-21: I see what you mean. I updated, is it better.

relay/src/handlers.rs
andrei-21: Good idea. I'll take a note of that and consider it the next time I touch the file.

relay/src/handlers.rs
```diff
@@ -236,6 +193,116 @@ fn update_cache_if_needed(state: &AppState, key: &CacheKey, signed_packet: &Sign
...
+
+    // DhtNetworkOnly reports the DHT's current state without using the relay
+    // cache as a lower bound or as invalid-sequence suppression.
+    let resolved = state.dht.resolve(public_key, None).await.complete().await;
+    log_resolve_warnings(public_key, &resolved.report, state.report_policy);
+
+    Ok(resolved.most_recent?)
+}
+
+fn apply_cached_floor(
```
SeverinAlexB, line 286: NIT: This method might benefit from a doc string as `apply_cached_floor` is hard to interpret

SeverinAlexB (approved):
One NIT, only change it if you think it's needed. I am approving already so you can merge it if you change or not.

PR #266, opened 2026-07-04
refactor(client)!: split networking backends

andrei-21:
Move the client implementation into focused modules for cache handling,
backend routing, relay fanout, result aggregation, and error mapping.

Reuse the same Rust client path from JS bindings and relay-only builds
so DHT, relay, and combined-network behavior share the same aggregation
rules.

BREAKING CHANGE: Client::publish now returns Result<u32, PublishError>.
The u32 is stored_on, the maximum DHT stored-node count reported by any
successful backend.

BREAKING CHANGE: Client::resolve now requires a ResolvePolicy argument
and returns Result<SignedPacket, ResolveError>. The previous resolve()
optional-return behavior is now ResolvePolicy::CacheFirst, and
resolve_most_recent() is now ResolvePolicy::DhtNetworkOnly.

BREAKING CHANGE: ClientBlocking, Client::cache(), Client::dht(), the
public client::relays module, and the previous client futures helpers
were removed. Use async Client, RelayClient, and ResolvePolicy instead.

# Review Guide

### `pkarr/src/client/mod.rs`

The client module root is the public surface and module index.

### `pkarr/src/client/client_impl.rs`

The high-level Pkarr client handles local concerns:

- owns cache configuration and cache reads/writes;
- applies `ResolvePolicy` behavior before and after backend calls;
- wraps public async operations for runtime compatibility;
- routes network work to `pkarr/src/client/backend/`.

### `pkarr/src/client/backend/mod.rs`

The backend module routes requests to the configured network implementation:

- DHT only;
- relays only;
- both DHT and relays.

### `pkarr/src/client/backend/dht.rs`

The DHT backend contains DHT-specific resolve behavior and publish warning
logging.

### `pkarr/src/client/backend/both.rs`

The combined backend runs relays and DHT together.

### `pkarr/src/client/backend/relays.rs`

The relays client fans requests out to multiple relay endpoints.

### `pkarr/src/client/backend/publish_results.rs`

Publish result aggregation is shared by relay-only and combined-network
publishes.

### `pkarr/src/client/backend/resolve_results.rs`

Resolve result aggregation is shared by relay-only and combined-network
resolves.

### `pkarr/src/relay_client.rs`

The relay client talks to one relay instance.


docs/integration.md
```diff
@@ -142,43 +142,23 @@ let client = Client::builder()
     .build()?;
```
SeverinAlexB, line 167: Is the feature flag decision tree still up to date in this doc?

SeverinAlexB: This doc seems to be a bit out of date.

AI review so take it with the required grain of salt:

Image

Image

docs/integration.md
```diff
@@ -228,10 +210,11 @@ async fn republish_loop(
 
 ### Safe Republishing with CAS
 
-When multiple processes might update the same key, use CAS (compare-and-swap) to prevent lost updates:
+When multiple processes might update the same key, use CAS (compare-and-swap) to prevent lost updates.
```
SeverinAlexB: NIT

```suggestion
When multiple processes update the same key, use CAS (compare-and-swap) to prevent lost updates.
```

docs/integration.md
```diff
@@ -240,63 +223,41 @@ async fn safe_republish(
...
                     .sign(keypair)?;
                 (new_packet, None)
             }
+            Err(error) => return Err(error.into()),
         };
 
     // Publish with CAS - fails if someone else published in between
-    client.publish(&current_packet, cas).await?;
+    let stored_on = client.publish(&current_packet, cas).await?;
+    println!("Published successfully; stored on at least {stored_on} DHT nodes");
```
SeverinAlexB: Out of scope for this PR but I am wondering, because we are already doing a big breaking change, should a publish with CAS be its own method? This would simplify the API for newcomers.

```rust
// Easy method that everybody instantly understands
client.publish(&packet).await?;

// Advanced method for CAS
client.compare_and_swap(&packet, previous_timestamp).await?;
```

This feels more intuitive.

docs/integration.md
```diff
@@ -240,63 +223,41 @@ async fn safe_republish(
...
-    // Check if bootstrapped
-    let bootstrapped = dht.bootstrapped();
-
-    // Get routing table info
-    let info = dht.info();
-
-    // Export bootstrap nodes for other clients
-    let bootstrap_nodes = dht.to_bootstrap();
-}
-```
```
SeverinAlexB, line 286: How get access this information now?

docs/quickstart.md
```diff
@@ -78,19 +78,19 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     println!("Publishing {} ...", keypair.public_key());
 
-    client.publish(&signed_packet, None).await?;
+    let stored_on = client.publish(&signed_packet, None).await?;
 
-    println!("Published successfully!");
+    println!("Published successfully; stored on at least {stored_on} DHT nodes");
```
SeverinAlexB, line 83: Conceptually, is it stored on "at least" x nodes or "on" x nodes?

You write:

>The returned `stored_on` value is the maximum count reported by any successful backend, not a sum, because backends may store the packet on the same DHT nodes. It means the packet was stored on at least that many DHT nodes.

Here you say it is "the maximum"?

pkarr/src/client/backend/both.rs
```diff
@@ -0,0 +1,121 @@
...
+use super::resolve_results::ResolveResults;
+use super::ResolveError;
+use crate::client::PublishError;
+
+pub(super) async fn publish(
+    relays: &RelaysClient,
+    dht: &DhtBackend,
+    signed_packet: &SignedPacket,
+    cas: Option<Timestamp>,
+) -> Result<u32, PublishError> {
```
SeverinAlexB: Should these methods be combined into a DhtRelaysBackend? Both methods take relays/dht args so it hints towards this.

pkarr/src/client/backend/both.rs
```diff
@@ -0,0 +1,121 @@
...
+    let relay_publish = relays.publish(signed_packet, cas);
+    let dht_publish = dht.publish(signed_packet, cas);
+    let (relay_result, dht_result) = tokio::join!(relay_publish, dht_publish);
```
SeverinAlexB: General question: You are using tokio::join here. So far, the lib has been async runtime independent as far as I know. Will using tokio here make it dependent on tokio and incompatible with other asycn rutimes?

Image

AI answer:
Image

pkarr/src/client/backend/resolve_results.rs
```diff
@@ -0,0 +1,71 @@
+use std::collections::HashMap;
+
+use crate::client::ResolveError;
+use crate::SignedPacket;
+
+#[derive(Default)]
+pub(super) struct ResolveResults {
```
SeverinAlexB: This needs a doc string. What is this for? What is the record method doing?

Also, usually when the impression of "need docstring" comes up, the naming choice of the struct/methods is often bad. One can do a lot by choosing good names so actual docstrings are not needed.

pkarr/src/client/backend/publish_results.rs
```diff
@@ -0,0 +1,114 @@
+use std::collections::HashMap;
+
+use crate::client::{ConcurrencyError, PublishError};
+
+#[derive(Default)]
+pub(super) struct PublishResults {
```
SeverinAlexB: Doc string needed

pkarr/src/client/backend/dht.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    pub(super) async fn publish(
+        &self,
+        signed_packet: &SignedPacket,
+        cas: Option<Timestamp>,
+    ) -> Result<u32, PublishError> {
+        let stored_on = self.client.publish(signed_packet, cas).await?;
+        self.log_publish_warnings(&signed_packet.public_key(), stored_on);
+        Ok(stored_on)
+    }
```
SeverinAlexB, line 31: Idea: You are returning the stored_on u32 number in a lot of methods. As mentioned before, looking at any `publish() -> u32` method from the outside, it's hard to interpret what `u32` means.

Should we use a type alias to make the meaning of u32 explicit?

```rust
/// Number of nodes the publish has stored the packet on.
pub type NumNodesStoredOn = u32;

fn publish(...) -> Result<NumNodesStoredOn, PublishError> {}
```

This relieves you from making doc strings for ever publish method. It's still a u32 at the end so it can be used as usual.

pkarr/src/client/backend/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    pub(super) fn try_merge(self, other: Backend) -> Option<Self> {
+        match (self, other) {
+            #[cfg(all(dht, relays))]
+            (Backend::Dht(dht), Backend::Relays(relays))
+            | (Backend::Relays(relays), Backend::Dht(dht)) => Some(Backend::Both(relays, dht)),
+            _ => None,
+        }
+    }
```
SeverinAlexB: "try" indicates this method can fail. This method does not return a `Result` though. Something feels off with the method too but I am not sure what it is.

Edit: A merge function should probably support any merge:
- self: relays, other: relays => merge relays into a single one
- self: dht, other: relays => both (already implemented)
- self: relays, other: dht => both (already implemented)
- self: dht, other: dht => Error

pkarr/src/client/backend/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+#[cfg(relays)]
+mod resolve_results;
+
+use super::errors::ResolveError;
+use super::{BuildError, PublishError};
+#[cfg(relays)]
+use relays::RelaysClient;
+
+#[derive(Debug)]
+pub(super) enum Backend {
```
SeverinAlexB: Please use the `mod.rs` files only for im/exports, not actual code. This should live in it's own file. Almost missed reviewing it.

pkarr/src/client/client_impl.rs
```diff
@@ -0,0 +1,286 @@
...
+    minimum_ttl: u32,
+    maximum_ttl: u32,
+    cache: Option<Arc<dyn Cache>>,
+    backend: Arc<Backend>,
+    #[cfg(feature = "endpoints")]
+    pub(crate) max_recursion_depth: u8,
+}
+
+impl Client {
+    /// Returns a builder to edit config before creating Client.
```
SeverinAlexB:
```suggestion
    /// Returns a builder for editing the config before creating a `Client`.
```

pkarr/src/client/errors.rs
```diff
@@ -0,0 +1,240 @@
+#[derive(thiserror::Error, Debug)]
+/// Errors occurring during building a [Client][super::Client]
+pub enum BuildError {
+    #[error("Client configured without Mainline node or relays.")]
+    /// Client configured without Mainline node or relays.
+    NoNetwork,
```
SeverinAlexB: Is "Mainline" the right word here? I think you meant DHT, no?

pkarr/src/client/errors.rs
```diff
@@ -0,0 +1,240 @@
...
+    DhtBuildError(std::io::Error),
+
+    #[cfg(relays)]
+    #[error("Failed to build the relays client {0}")]
+    /// Failed to build the relays client.
+    RelayBuildError(crate::relay_client::RelayError),
+
+    #[error("Passed an empty list of relays")]
+    /// Passed an empty list of relays
+    EmptyListOfRelays,
```
SeverinAlexB: Shouldn't this be a `RelayBuildError` too?

pkarr/src/client/errors.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+#[derive(thiserror::Error, Debug, Clone, PartialEq, Eq, Hash)]
+/// Errors occurring during publishing a [SignedPacket][crate::SignedPacket]
+pub enum PublishError {
+    #[error(transparent)]
+    /// Errors that requires either a retry or debugging the network condition.
+    Query(#[from] QueryError),
```
SeverinAlexB: If a publish can have a "QueryError", isn't "QueryError" the wrong name?

pkarr/src/client/errors.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Errors that require either a retry or debugging the network condition.
+    Query(#[from] QueryError),
+
+    #[error("DHT resolve queried no nodes")]
+    /// DHT resolve queried no nodes.
+    NoDhtNodesQueried,
+
+    #[error("DHT resolve found no valid responses")]
+    /// DHT resolve received responses, but none contained valid mutable records.
+    NoValidDhtResponses,
```
SeverinAlexB: QueryError already covers similar errors like NoDht and NoValidDhtResponse. So how are they different?

```
/// Errors that requires either a retry or debugging the network condition.
pub enum QueryError {
    /// Publish query timed out with no responses neither success or errors, from Dht or relays.
    #[error("Publish query timed out with no responses neither success or errors.")]
    Timeout,

    #[error("Publishing SignedPacket to Mainline failed.")]
    /// Publishing SignedPacket to Mainline failed.
    NoClosestNodes,

    #[error("Publishing SignedPacket to Mainline failed code: {0}, description: {1}.")]
    /// Publishing SignedPacket to Mainline failed, received an error response.
    DhtErrorResponse(i32, String),

    #[error("Most relays responded with bad request")]
    /// Most relays responded with bad request
    BadRequest,
}
```

SeverinAlexB (changes requested):
Good change in general

pkarr/src/client/backend/dht.rs
andrei-21: Nice idea, I added `StoredNodeCount` type alias.

pkarr/src/client/backend/mod.rs
andrei-21: Moved to `backend_impl.rs`.

pkarr/src/client/backend/mod.rs
andrei-21: Renamed to `checked_merge()`. The standard library use `checked_` prefix in such cases, e.g. or for arithmetic-style ops.
I left the implementation the same. Even if `self: relays, other: relays` case may exist, in practice it does not make sense, I added a doc explaining this.

pkarr/src/client/errors.rs
andrei-21: Right, changed to `DHT client` as in other places below.

pkarr/src/client/errors.rs
andrei-21: No, `EmptyListOfRelays` is an error when building a relays client (note plural); `RelayBuildError` is an error when building a relay client (note singular).

pkarr/src/client/client_impl.rs
andrei-21: Fixed.

pkarr/src/client/backend/both.rs
andrei-21: Done.

pkarr/src/client/backend/both.rs
andrei-21: Well, I just added usage of a Tokio macro, while the Tokio dependency was already there. So the implementation remains runtime-agnostic.
However, if we want to get rid of the Tokio dependency, then I made it a bit worse.
I will leave it as is. If we later decide to remove Tokio, we can easily replace `tokio::join!()`, but to replace `tokio::select!()` we will need an ugly function to poll futures manually.
Note we will still need `async_compact` because of `reqwest`.

docs/quickstart.md
andrei-21: Backend 1 can publish on 3 nodes: A, B, C.
backend 2 can publish on 2 nodes: C, D.
We can be sure that we published at least on 3 nodes, but maybe more if different relays/backends published on different nodes.

docs/integration.md
andrei-21: We don't. Why do we need this? In `pubky-core` it's used in `pubky-homeserver/src/republishers/pkarr_republisher/verify.rs`, which will not be needed with the new `publish()` method that returns `stored_on`.

docs/integration.md
andrei-21: Hm. I am not sure that it is a sane idea to use CAS in a distributed network, especially in the case of using multiple backends.
Even the spec is not precise enough:
> If the sequence number is equal, and the value is also the same, the node SHOULD reset its timeout counter.
> ...
> Only if the cas matches the stored sequence number is the put performed. If it mismatches, the store fails and an error MUST be returned.

If we want to properly implement it, we need to research actual implementations of DHT nodes, their behavior, and design a reliable, predictable feature.

docs/integration.md
andrei-21: Done.

docs/integration.md
andrei-21: Fixed.

pkarr/src/client/backend/resolve_results.rs
andrei-21: Added a doc string, changed naming to *Accumulator*.

pkarr/src/client/backend/publish_results.rs
andrei-21: Added a doc string, changed naming to *Accumulator*.

pkarr/src/client/errors.rs
andrei-21: Well, yes, it was inherited...
Anyway I made a major refactoring of error enums, now `QueryError` is gone.

pkarr/src/client/errors.rs
andrei-21: Well, yes, it was inherited...
Anyway I made a major refactoring of error enums, now `QueryError` is gone.

docs/integration.md
SeverinAlexB: Interesting. Do I understand this correct: You are questioning CAS in general, not the publish simlification proposal?

So should we just remove CAS then?

docs/integration.md
SeverinAlexB: Wasn't the `dht.info()` property used to check the DHT size estimation? Not sure but I think so

docs/integration.md
andrei-21: Yes I am questioning CAS in general. Well... it is used in homeserver... not sure if it should. So I decided to keep the status quo.

pkarr/src/client/backend/both.rs
SeverinAlexB: Ok cool, let's push this out as I assume it will not be a problem for existing pkarr users. Maybe should remove tokio in the future if possible. I wonder how much bytes is added to the pkarr lib due to the tokio dependency.

> Note we will still need async_compact because of reqwest.

async_compact makes a lot of sense.

docs/integration.md
andrei-21: Not in `pubky-core`, as far as I can see. Maybe you mean `pkarr-relay`, where it is used, but it is already in `pkarr`.

pkarr/src/client/backend/both.rs
andrei-21: With `1746044` bytes, without `1705768`.

docs/integration.md
SeverinAlexB: Happy to remove it and simplify the publish method. We can always add it again later. I don't think it's such an important feature. Up to you how/if you want to do it though.

pkarr/src/client/backend/both.rs
SeverinAlexB: Sounds like a minor optimization. You decide if its worth it

andrei-21 (commented):
I removed CAS functionality.
Fixed some edge cases with packets received from relays but considered expired by the local TTLs. For that, `CacheContext` was introduced to encapsulate such logic and `BackendResolvePolicy` to nicely pass `CacheContext` when necessary.

bindings/js/src/client.rs
```diff
@@ -99,7 +94,8 @@ impl Client {
     ) -> Result<Option<super::SignedPacket>, JsValue> {
         // TODO: This could implement more sophisticated logic to
```
86667, line 95: I guess this TODO can be removed now

docs/integration.md
```diff
@@ -142,43 +167,23 @@ let client = Client::builder()
...
-
-    let packet = client.resolve(&public_key);
+    let packet = client.resolve(&public_key, ResolvePolicy::CacheFirst).await;
     Ok(())
 }
 ```
 
-### Runtime Agnostic
-
 Pkarr uses `async_compat` internally. If no Tokio runtime is detected, it automatically wraps futures for compatibility. This means pkarr works with async-std, smol, or any other executor.
```
86667, line 187: nit: if i understood previous comments about tokio then this should be removed

pkarr/src/types/resolve_policy.rs
```diff
@@ -16,31 +16,38 @@ pub enum ResolvePolicy {
     /// Return only a locally cached or relay-cached packet, even if expired.
     ///
     /// This is guaranteed to be fast and does not touch DHT nodes.
-    /// Useful for republishing.
+    /// Relay results populate the local cache. Useful for republishing.
     LocalOrRelayCacheOnly,
```
86667: I'm not a fan of this name. Why does the location of the cache need to be mentioned in the name when the policy difference is the behaviour after reading from cache? 
We either return only from cache, from cache or if expired then network, or only network. 

So `CacheFirst`, `CacheOnly`, `DhtNetworkOnly` is fine?

pkarr/README.md
```diff
@@ -12,7 +12,6 @@ Rust implementation of [Pkarr](https://github.com/pubky/pkarr) for publishing an
 ### Runtime Support
```
86667, line 12: Would you mind updating this readme with a little about the structure of the abstractions you've introduced here

pkarr/src/client/backend/resolve_policy.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Resolve only from local or relay caches.
+    LocalOrRelayCacheOnly,
+    /// Apply a cached packet floor and local TTL bounds to network responses.
+    CacheFirst(CacheContext<'a>),
+    /// Query configured backends for DHT network state without cached packet state.
+    DhtNetworkOnly,
+}
+
+#[cfg(relays)]
+impl<'a> BackendResolvePolicy<'a> {
```
86667: I'm not sure i understand why this is gated to the `relays` backend, whereas dht backend destructures the type itself

86667: Maybe the logic implemented here really lives with or on `RelayBackend`. That would remove this awkward backend-specific logic on a shared type.

86667 (commented):
Nice code.

bindings/js/src/client.rs
andrei-21: I haven't really refactored the SDK yet; I made just minimal changes. I will properly address it in the following PRs.

docs/integration.md
andrei-21: No, it still holds; I didn't really change anything in terms of behavior here.

pkarr/src/client/backend/resolve_policy.rs
andrei-21: Fair. I moved it to `relay.rs` and encapsulated it in `RelayResolveOptions`.

pkarr/src/types/resolve_policy.rs
andrei-21: I do not like the long name either. However, I am afraid that `CacheOnly` may feel like it accesses only the local cache. On the other hand, if I do not have relays configured, then it also feels awkward.
I would rename to `CacheOnly`, `CacheFirst`, and maybe even `NoCache`. @SeverinAlexB we already had some discussions about that; what do you think?

pkarr/README.md
andrei-21: I will, but if it's okay to do so in subsequent PRs, I'm trying to focus only on `pkarr::client` here—it's already getting big and complicated.

86667: Of course. Other things to update which i noticed are:
- Mention `relay` binary in root readme
- Update or remove CHANGELOGs

andrei-21: Noted.

pkarr/src/types/resolve_policy.rs
andrei-21: As discussed renamed to `CacheOnly`, `CacheFirst`, and `NetworkOnly`.

docs/integration.md
SeverinAlexB: So far, one could access the DHT size estimation in pkarr to do some sanity checks:

```rust
let client = pkarr::Client::builder().build()?;

if let Some(dht) = client.dht() {
    let (estimated_nodes, standard_deviation) =
        dht.info().dht_size_estimate();

    println!(
        "Estimated DHT size: {estimated_nodes}, standard deviation: {:.1}%",
        standard_deviation * 100.0
    );
}
```

This is not possible anymore because `client.dht()` is being removed. I am not saying we necessarily needs this in the pkarr client but it was very useful to determine if the pkarr connection was stable.

What's the purpose in hidding the `dht()`?

docs/quickstart.md
SeverinAlexB: Ahh now I get it. This is hard to understand when just reading the docs.

Can we do something like this?

> The returned stored_on value presents the minimum number of nodes your packet has been stored on. It is calculated by the maximum count reported by any successful backend, not a sum, because backends may store the packet on the same DHT nodes.

This gives the quick summery first (stored on at least x nodes) and then goes into detail in how it is calculated for interested devs. What do you think?

86667 (approved):
Im overall happy with these changes, they are certainly a large improvement. 

A couple things to follow-up on:
- As mentioned elsewhere - update readmes and remove or update changelogs
- The `Cache` logic could be centralised more i think. In particular `update_cache_if_needed()` living in service layer feels off. Also, could we bring cache trait and its implementations into a single place?

docs/integration.md
andrei-21: Well, the purpose in hiding the `dht()` is to avoid exposing implementation details. On the other hand, I agree that letting clients determine whether the pkarr connection was stable might be a good idea. I would say we need, first, to make pkarr as reliable as possible; second, to find the right abstractions to let clients learn how healthy the pkarr connection is without knowing details of the DHT network.
Answering your question: I removed it because I do not feel that it is the right abstraction; I didn't propose anything instead to keep the scope of the PR smaller.

docs/quickstart.md
andrei-21: Updated.

SeverinAlexB (approved):

PR #267, opened 2026-07-05
refactor(relay): split DHT service and index rendering

andrei-21:
Move cache-aware DHT publish and resolve logic out of HTTP handlers
into a relay DHT service. Move index page rendering into its own
module.

### DIFF
--- a/relay/src/dht_service.rs
+++ b/relay/src/dht_service.rs
@@ -1,0 +1,377 @@
+use std::sync::{Arc, Mutex};
+
+use http::StatusCode;
+use pkarr::dht::{
+    DhtClient, DhtInfo, ReportPolicy, ResolveError as DhtResolveError, ResolveReport,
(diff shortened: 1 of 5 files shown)


relay/src/dht_service.rs
```diff
@@ -0,0 +1,377 @@
...
+            report_policy,
+            rate_limiter,
+        }
+    }
+
+    /// Publish a signed packet through the DHT and update the relay cache.
+    pub(crate) async fn publish(
+        &self,
+        public_key: &PublicKey,
+        signed_packet: &SignedPacket,
```
SeverinAlexB, line 54: Public key is redundant here as it can be retrieved with `signed_packet.public_key()`

SeverinAlexB (approved):
Beautiful. One little thing, otherwise it's good to go. I approved already so you don't need to request a review again.

PR #268, opened 2026-07-06
feat(relays)!: return stored node count from publish

andrei-21:
Add Pkarr-Dht-Stored-Nodes to successful relay PUT responses and
expose it through CORS. Parse the header in RelayClient::publish,
defaulting to 1 for old relays that do not return it.

BREAKING CHANGE: RelayClient::publish now returns Result<u32, RelayError>
instead of Result<(), RelayError>.

### DIFF
--- a/pkarr/src/relay_client.rs
+++ b/pkarr/src/relay_client.rs
@@ -9,7 +9,9 @@ use reqwest::{
 use std::time::Duration;
 use url::Url;
 
-use crate::{PublicKey, ResolvePolicy, SignedPacket, PKARR_INVALID_SIGNED_PACKET_SEQ};
+use crate::{
+    PublicKey, ResolvePolicy, SignedPacket, PKARR_DHT_STORED_NODES, PKARR_INVALID_SIGNED_PACKET_SEQ,
+};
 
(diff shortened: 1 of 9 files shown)


pkarr/src/relay_client.rs
```diff
@@ -87,8 +96,9 @@ impl RelayClient {
         let status = response.status();
 
         if status.is_success() {
+            let stored_at = extract_dht_stored_nodes(&response)?;
             debug!("Successfully published to {url}");
-            return Ok(());
+            return Ok(stored_at);
```
SeverinAlexB: The `_at` suffix usually indicates a timestamp or datetime. Can we rename it to `stored_on` or so?

This comment obviously is not just for the marked code but for the whole codebase. I hope I am not noting this too late.

SeverinAlexB (changes requested):
One thing, otherwise good.

pkarr/src/relay_client.rs
andrei-21: Fixed.

SeverinAlexB (approved):

PR #269, opened 2026-07-15
feat(js)!: align bindings with client API

andrei-21:
Expose resolve policies and structured errors. Return the stored node
count from publish and use the core request timeout by default.

Normalize JavaScript naming and timestamp units, harden generated ESM
exports, and update documentation and tests for the revised API.

BREAKING CHANGE: resolve now requires a ResolvePolicy and rejects when
no packet is found. Remove resolveMostRecent, use camelCase Keypair
methods, return milliseconds from timestampMs, and remove isValid and
Utils.defaultRelays.


bindings/js/LICENSE
```diff
@@ -0,0 +1 @@
+../../LICENSE
```
SeverinAlexB: AI review but I think this could be correct:

Medium: The published npm package loses its license file. bindings/js/pkg/LICENSE:1 is now a symlink to another symlink, while bindings/js/pkg/package.json:10-16 does not explicitly include it. npm excludes symlinks from package archives, so the previous regular LICENCE file is no longer shipped. Keep a regular LICENSE inside pkg or copy it there during packaging.

bindings/js/pkg/README.md
```diff
@@ -7,13 +7,13 @@ Public-Key Addressable Resource Records for publishing and resolving DNS packets
 This package is generated via `wasm-pack` and includes TypeScript definitions.
 
 ```bash
-npm install pkarr
+npm install @synonymdev/pkarr
```
SeverinAlexB, line 10: Nice catch

bindings/js/pkg/README.md
```diff
@@ -26,10 +26,14 @@ builder.addARecord("www", "192.168.1.1", 3600);
...
 // Sign and publish
 const packet = builder.buildAndSign(keypair);
-await client.publish(packet);
+const storedNodeCount = await client.publish(packet);
+console.log(`Stored on at least ${storedNodeCount} DHT nodes`);
 
 // Resolve later
-const resolved = await client.resolve(keypair.public_key_string());
+const resolved = await client.resolve(
+    keypair.publicKeyString(),
```
SeverinAlexB: I wonder, should we even name this `publicKey()` directly without the "String". String is the default representation of a public key here as far as I can see?

bindings/js/pkg/README.md
```diff
@@ -38,43 +42,44 @@ console.log('Records:', resolved.records);
...
 #### Client Methods
 
 ```javascript
-const { Client } = require('pkarr');
+const { Client, ResolvePolicy } = require('@synonymdev/pkarr');
 
-const client = new Client();                    // Default relays
-const client = new Client(relays, timeout);     // Custom configuration
+const client = new Client();                          // Default relays and 2s timeout
+const customClient = new Client(relays, timeout);     // Custom configuration
```
SeverinAlexB: Not sure what we decided but a 2s relay request timeout feels very low especially when somebody is in a country with bad mobile internet, think El Salvador or so.

bindings/js/pkg/README.md
```diff
@@ -108,7 +113,9 @@ console.log(packet.timestampMs);
 console.log(packet.records);
 ```
 
-CAS ensures your update only succeeds if the server state hasn't changed since you last read it.
+`resolve` rejects when no packet is found or the configured relays fail. Pkarr errors are
+JavaScript `Error` objects with a stable `code`, such as `NotFound`, `NoResponses`, or
+`NotMostRecent`.
```
SeverinAlexB: Hmmm, I think in JS, we need to return `null` if no packet is found. Interpreting JS error is awkward so I think returning `null` in this case is easier.

SeverinAlexB (changes requested):
Early review. I will need to do a second round later. Overall LGTM

bindings/js/pkg/README.md
andrei-21: Sev congratulated me (meanwhile I was just accepting what AI suggested).
Image

bindings/js/LICENSE
andrei-21: Fair. Fixed.

bindings/js/pkg/README.md
andrei-21: Hm... we are back to the `Result<Option<SignedPacket>>` problem.
What if we add `isNoValidPacketError(error: unknown): boolean` function which returns true for `NotFound` and `InvalidSignedPacket`?
Then usage:
```js
try {
    return await client.resolve(publicKey, policy);
} catch (error) {
    if (isNoValidPacketError(error)) {
        return null;
    }

    throw error;
}
```

bindings/js/pkg/README.md
andrei-21: Makes sense. Renamed.

bindings/js/pkg/README.md
andrei-21: It is just a default value matching what we have in the pkarr client; the user can always customize it if needed. If we want to change the default, then I think we'd better change it in the pkarr client.

SeverinAlexB: I think we need to increase the relay request timeout in general. 2s is just not enough for it to be reliable on mobile. Maybe 5s? Or do you have an other idea to make this work?

SeverinAlexB:
When running this JS code with this PR, I get an error:

```js
  const client = new Client();
  const keypair = new Keypair();
  const publicKey = keypair.publicKeyString();
  const builder = SignedPacket.builder();
  builder.addTxtRecord("_hello", "world", 60);
  const packet = builder.buildAndSign(keypair);

  console.log(`Publishing ${publicKey}...`);
  let startedAt = performance.now();
  const storedOn = await client.publish(packet);
  console.log(
    `Stored on at least ${storedOn} DHT nodes in ${elapsedMs(startedAt)} ms`,
  );
  ```
  
  ```bash
  ➜  js (master) npm run start                                                                                                     ✱

> pkarr-js-test@0.1.0 start
> node index.js

Publishing ad7f3xc39e1omn6duuuse5ad4jiu8c6p1t3nac54ytuszw1b6byy...
Error [PkarrPublishError]: publish query received no responses
    at __wbg_new_b667d279fd5aa943 (/Users/severinbuhler/git/pubky/pkarr/bindings/js/pkg/index.cjs:813:25)
    at wasm://wasm/002b8f6e:wasm-function[407]:0x69097
    at wasm://wasm/002b8f6e:wasm-function[122]:0x27574
    at wasm://wasm/002b8f6e:wasm-function[290]:0x5ccd2
    at wasm://wasm/002b8f6e:wasm-function[661]:0x83ab1
    at wasm://wasm/002b8f6e:wasm-function[542]:0x6feb3
    at wasm://wasm/002b8f6e:wasm-function[648]:0x81f36
    at wasm_bindgen__convert__closures_____invoke__hde909c3c6ef87133 (/Users/severinbuhler/git/pubky/pkarr/bindings/js/pkg/index.cjs:1044:22)
    at real (/Users/severinbuhler/git/pubky/pkarr/bindings/js/pkg/index.cjs:1208:20)
    at node:internal/process/task_queues:151:7 {
  code: 'NoResponses'
}
```

The js should only use relays so this is weird.

bindings/js/pkg/README.md
andrei-21: Oh, you are right; AI confused me.
So since JS uses only relays backend, this timeout is simply reqwest timeout. The previous 30-second timeout is very reasonable. The issue with increasing DHT timeout causing all resolves to take longer is not applicable here.
Reverted back and commented so that we do not confuse in the future.

andrei-21:
> When running this JS code with this PR, I get an error:
> code: 'NoResponses'
> The js should only use relays so this is weird.

No, that is expected. It means that the client received no responses from their backends (from no relays in JS case).

bindings/js/pkg/README.md
andrei-21: Change to return `null` on `NotFound`.

SeverinAlexB (approved):
LGTM

PR #270, opened 2026-07-15
build(cargo)!: simplify features and dependencies

andrei-21:
Make key utilities unconditional so key-only builds need no explicit
feature. Remove unused dependencies from pkarr, relay, and the JS
bindings, and warn when new unused crate dependencies are introduced.

Declare the Tokio features used directly by pkarr and relay, and align
the feature documentation with the simplified configuration.

BREAKING CHANGE: Remove the pkarr `keys` Cargo feature. Consumers using
`features = ["keys"]` must remove it; key utilities remain available
with `default-features = false`.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -13,23 +13,26 @@ pkarr = "6"
 
 This includes both DHT and relay support, suitable for most server applications.
 
-## Core Features
+## Base API
 
-### `keys`
+### Key utilities
 
-Ed25519 key utilities for identity management.
+Ed25519 key utilities for identity management are always available, including
+when default features are disabled.
 
 ```toml
-pkarr = { version = "6", default-features = false, features = ["keys"] }
+pkarr = { version = "6", default-features = false }
 ```
 
 Provides:
 - `Keypair` - Generate and manage Ed25519 keypairs
 - `PublicKey` - Handle public keys and derive z-base32 identifiers
 
+## Core Features
+
 ### `signed_packet`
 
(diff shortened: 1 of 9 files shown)


docs/features.md
```diff
@@ -13,23 +13,26 @@ pkarr = "6"
...
-### `keys`
+### Key utilities
 
-Ed25519 key utilities for identity management.
+Ed25519 key utilities for identity management are always available, including
+when default features are disabled.
 
 ```toml
-pkarr = { version = "6", default-features = false, features = ["keys"] }
+pkarr = { version = "6", default-features = false }
```
SeverinAlexB, line 24: AI review, the problem already existed before as far as I know. Up to you if you want to fix it.

1. Medium: Key-only rustdoc fails due to feature-gated links  
pkarr/src/types/keys.rs:15 and pkarr/src/types/keys.rs:117 link to crate::SignedPacket, but that type is unavailable without the signed_packet feature (pkarr/src/types/mod.rs:10-11). Since pkarr/src/lib.rs:8 denies broken intra-doc links, the newly documented minimal configuration cannot generate documentation:
cargo doc -p pkarr --lib --no-default-features --no-deps
Make these links conditional or use plain text when signed_packet is disabled.

SeverinAlexB (approved):

docs/features.md
andrei-21: Yes... it is a part of a bigger problem. I filed an issue for that https://github.com/pubky/pkarr/issues/271.

PR #273, opened 2026-07-16
docs: refresh client and builder guidance

andrei-21:
Update publish and resolve examples for current APIs, document resolve
policies, errors, backend configuration, and coordinated republishing.

Remove obsolete crate and relay changelogs.


docs/integration.md
```diff
@@ -75,6 +75,12 @@ let client = Client::builder()
     .relays(&["https://pkarr.pubky.app", "https://my-relay.example.com"])?
```
SeverinAlexB, line 75: This guides still references pkarr version 6 while using the new publish method. Probably should be bumped to version 7.

docs/integration.md
```diff
@@ -234,34 +289,52 @@ async fn republish_loop(
...
+use pkarr::{
+    errors::ResolveError, Client, Keypair, ResolvePolicy, SignedPacket, Timestamp,
+};
+use std::io;
+use tokio::sync::Semaphore;
 
 async fn coordinated_republish(
     client: &Client,
     keypair: &Keypair,
+    publish_gate: &Semaphore,
```
SeverinAlexB: Do we really need to complicate this example with a Semaphore? Sounds like an overkill for this feature illustration. Just mentioning it above feels enough.

docs/integration.md
```diff
@@ -234,34 +289,52 @@ async fn republish_loop(
...
+        .checked_add(1)
+        .filter(|sequence| *sequence <= i64::MAX as u64)
+        .ok_or_else(|| io::Error::other("Pkarr sequence is exhausted"))?;
+    let timestamp = Timestamp::from(Timestamp::now().as_u64().max(next_sequence));
+
+    let packet = SignedPacket::builder()
+        // Build the complete desired record set; omitted records are removed.
+        .txt("key".try_into()?, "updated_value".try_into()?, 300)
+        .timestamp(timestamp)
+        .sign(keypair)?;
```
SeverinAlexB, line 328: Is setting the timestamp really needed? As far as I know, pkarr will just set the current timestamp so managing the timestamp manually is an overkill for 99% of the projects.

docs/quickstart.md
```diff
@@ -116,7 +116,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             }
         }
         Err(pkarr::errors::ResolveError::NotFound) => {
```
SeverinAlexB, line 118: On line 104, there is still a `pk:` reference in the guide. Can you remove it and just use the plain key?

*Not your mistake but would be nice if we could fix it here.

docs/quickstart.md
```diff
@@ -127,7 +130,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 }
 ```
 
-Use `resolve(&public_key, ResolvePolicy::NetworkOnly)` when you need the latest version, for example before publishing updates. `ResolvePolicy::CacheFirst` only returns non-expired cached packets; use `ResolvePolicy::CacheOnly` when expired cached packets are acceptable.
+Use `resolve(&public_key, ResolvePolicy::NetworkOnly)` when you need the latest
+network state, for example before publishing updates. A newer mutable item that
+is not a valid Pkarr packet is returned as `ResolveError::InvalidSignedPacket`,
```
SeverinAlexB: We should always use PKARR with all capital letters.

```suggestion
is not a valid PKARR packet is returned as `ResolveError::InvalidSignedPacket`,
```

docs/quickstart.md
```diff
@@ -138,8 +149,9 @@ use pkarr::{Client, Keypair, ResolvePolicy, SignedPacket};
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    // 1. Create client and keypair
-    let client = Client::builder().build()?;
+    // 1. Use independent clients so resolving cannot read the publisher's cache
+    let publisher = Client::builder().build()?;
+    let resolver = Client::builder().cache_size(0).build()?;
```
SeverinAlexB: Instead of using two clients here, can't you just set cache_size to 0 and use one client? This would simplify the example

SeverinAlexB (changes requested):
A few comments, otherwise LGTM

docs/integration.md
andrei-21: I will update with the release.

docs/integration.md
andrei-21: To update we need to take max.

docs/integration.md
andrei-21: Fair, simplified.

docs/quickstart.md
andrei-21: Done.

docs/quickstart.md
andrei-21: Fixed.

docs/quickstart.md
andrei-21: Fixed.

SeverinAlexB (approved):

PR #274, opened 2026-07-17
chore: release v7.0.0

andrei-21:
## Changes since v6.0.1

- Add explicit `CacheOnly`, `CacheFirst`, and `NetworkOnly` resolve policies.
- Return structured publish and resolve errors, including invalid packet sequence information.
- Return the number of DHT nodes storing a packet from client and relay publish operations.
- Add DHT query diagnostics and hide direct `mainline` access behind the new `DhtClient` abstraction.
- Add the public single-endpoint `RelayClient`.
- Split relay HTTP, inbound DHT, and user-triggered DHT rate limits.
- Improve relay cache freshness handling, request timeouts, shutdown behavior, and Docker support.
- Align the JavaScript bindings with the Rust client API, including resolve policies, structured errors, stored-node counts, normalized naming, and millisecond timestamps.
- Make key utilities always available and remove the `keys` Cargo feature.
- Update `ed25519-dalek` to stable v3 and remove unused dependencies.

## Breaking changes

- `Client::resolve` now requires a `ResolvePolicy` and returns `Result<SignedPacket, ResolveError>`.
- `Client::publish` now returns the stored-node count and no longer accepts a CAS timestamp.
- `ClientBlocking`, `Client::cache`, `Client::dht`, the public client relay module, and the previous client future helpers were removed.
- `ClientBuilder::dht` now configures `mainline::Config`.
- Relay rate-limiter configuration now requires separate HTTP, DHT-peer, and user-triggered DHT quotas.
- `RelayBuilder::pkarr` was replaced by DHT and request-timeout configuration methods.
- The `keys` Cargo feature was removed; key utilities remain available with default features disabled.
- JavaScript client methods, key accessors, error handling, and timestamp fields were updated to match the new API.

### DIFF
--- a/bindings/js/pkg/package.json
+++ b/bindings/js/pkg/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@synonymdev/pkarr",
   "description": "Public-Key Addressable Resource Records (Pkarr); publish and resolve DNS records over Mainline DHT",
-  "version": "0.1.4",
+  "version": "1.0.0",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/relay/Cargo.toml
+++ b/relay/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pkarr-relay"
-version = "0.12.0"
+version = "1.0.0"
 rust-version = "1.85"
 authors = [
     "Nuh <nuh@nuh.dev>",
@@ -41,7 +41,7 @@ toml = "0.9.11"
 clap = { version = "4.5.57", features = ["derive"] }
 httpdate = "1.0.3"
 url = "2.5.8"
-pkarr = { path = "../pkarr", version = "6.0", default-features = false, features = [
+pkarr = { path = "../pkarr", version = "7.0.0", default-features = false, features = [
     "dht",
     "lmdb-cache",
 ] }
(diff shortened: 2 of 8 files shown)


bindings/js/pkg/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@synonymdev/pkarr",
   "description": "Public-Key Addressable Resource Records (Pkarr); publish and resolve DNS records over Mainline DHT",
-  "version": "0.1.4",
+  "version": "0.2.0",
```
SeverinAlexB: Should we bump to v1.0.0?

relay/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pkarr-relay"
-version = "0.12.0"
+version = "0.13.0"
```
SeverinAlexB: Should we do v1.0.0?

bindings/js/pkg/package.json
andrei-21: Done.

relay/Cargo.toml
andrei-21: Done.

andrei-21:
Also note that I synchronized `pkarr-js` version to `7.0.0`.

SeverinAlexB (approved):

PR #280, opened 2026-07-23
docs: document client and relay architecture

andrei-21:
Describe builder, cache, backend, and resolve-policy interactions.
Add relay binary usage, current CLI options, and default network binding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,17 @@ async fn main() -> anyhow::Result<()> {
 | **[Examples](./pkarr/examples/README.md)** | Code samples |
 | **[Specifications](./design/README.md)** | Protocol design documents |
 
+## Repository Components
+
+- **`pkarr`** — The Rust library for keys, signed DNS packets, caching,
+  publishing, and resolution.
+- **`pkarr-relay`** — An HTTP-to-DHT relay server for browsers and other
+  UDP-restricted clients. Install it with `cargo install pkarr-relay`, or run
+  the workspace binary with `cargo run -p pkarr-relay -- --help`. See the
+  [relay guide](./relay/README.md) for configuration and testnet usage.
+- **`@synonymdev/pkarr`** — JavaScript/WASM bindings under
+  [`bindings/js`](./bindings/js/README.md).
+
 ## Demo
 
 Try the [web app](https://pkdns.net) to resolve records in your browser.
@@ -78,6 +89,12 @@ sequenceDiagram
     Relay->>Client: Verified response
 ```
 
+The default native client configures both paths. `ClientBuilder` can instead
+select DHT-only or relay-only operation, and `ResolvePolicy` controls whether a
+lookup uses cached data or queries the configured networks. See the
+[integration guide](./docs/integration.md#client-abstractions) for the client
+structure and selection behavior.
+
 ### The Network
 
 PKARR uses the [Mainline DHT](https://en.wikipedia.org/wiki/Mainline_DHT), the same peer-to-peer network that powers BitTorrent. Records are stored using [BEP44](https://www.bittorrent.org/beps/bep_0044.html) (mutable items). With 15 years of proven reliability and 10+ million active nodes, there's no need to bootstrap a new network.
@@ -89,7 +106,8 @@ PKARR uses the [Mainline DHT](https://en.wikipedia.org/wiki/Mainline_DHT), the s
 - **Caching everywhere** — Clients and relays cache aggressively for performance
 - **Relays for browsers** — Web apps use HTTP relays since browsers cannot open UDP sockets
 
-PKARR is the I/O library that reads and writes DNS records to the DHT.
+The `pkarr` crate is the I/O library that reads and writes signed DNS packets
+through the DHT, HTTP relays, or both.
 
 ## FAQ
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@
 
 [![Crates.io](https://img.shields.io/crates/v/pkarr)](https://crates.io/crates/pkarr) [![Documentation](https://docs.rs/pkarr/badge.svg)](https://docs.rs/pkarr) [![License](https://img.shields.io/badge/license-MIT-purple)](./LICENSE)
 
-PKARR turns Ed25519 public keys into domain names that you truly own. Publish DNS records to the Bittorrent peer-to-peer network with 10+ million nodes. No registrar can seize your domain. No platform can deplatform your identity.
+PKARR turns Ed25519 public keys into domain names that you truly own. Publish
+DNS records to the Mainline DHT without relying on a registrar or platform.
 
 Where we are going, this https://o4dksfbqk85ogzdb5osziw6befigbuxmuxkuxq8434q89uj56uyy resolves everywhere!
 
@@ -12,13 +13,14 @@ Where we are going, this https://o4dksfbqk85ogzdb5osziw6befigbuxmuxkuxq8434q89uj
 
 ```bash
 cargo add pkarr
+cargo add tokio --features macros,rt-multi-thread
 ```
 
 ```rust
 use pkarr::{Client, Keypair, SignedPacket};
 
 #[tokio::main]
-async fn main() -> anyhow::Result<()> {
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Generate your identity
     let keypair = Keypair::random();
     println!("Your public key: {}", keypair.public_key());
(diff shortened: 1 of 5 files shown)


docs/integration.md
```diff
@@ -48,6 +48,50 @@ already enabled on native targets. If you disable default features, combine
...
+      Client
+      ├── publish(&SignedPacket)
+      └── resolve(&PublicKey, ResolvePolicy)
+```
+
+`ClientBuilder` owns configuration: network selection, cache choice, TTL
+bounds, request timeout, and backend-specific settings. `build()` validates
+that at least one network is available and constructs a cloneable `Client`.
+The concrete DHT, relay, and combined backend types are internal; use builder
+methods and Cargo features to select them.
```
SeverinAlexB: You use two words for the same thing: `network` and `backend`. Should we decide on one and stick to it? We probably should always talk about `backends` and not networks, shouldn't we?

docs/integration.md
```diff
@@ -48,6 +48,50 @@ already enabled on native targets. If you disable default features, combine
...
+
+`Client` is the public I/O facade. It coordinates the optional local cache with
+the configured backend and maps backend-specific outcomes into `BuildError`,
+`PublishError`, and `ResolveError`. The `Cache` trait is the extension point for
+custom storage, while `ResolvePolicy` controls how a particular lookup balances
+cache latency against network freshness.
```
SeverinAlexB: Hmmm, this is an "Integration" doc to help users integrate pkarr into their app. Here you talk about internal implementation details. It feels like this is not the right place to describe this, especially because it is the second chapter in this doc. These internal implementation details are great for contributors but not necessary for pkarr users.

pkarr/README.md
```diff
@@ -1,23 +1,59 @@
...
+  `CacheOnly` avoids DHT queries, `CacheFirst` returns the fastest fresh result
+  without going backward from an expired cached packet, and `NetworkOnly`
+  aggregates the configured networks for their most recent observed state.
+
+The DHT, relay, and combined backend implementations are internal details.
+Select them through `ClientBuilder` rather than depending on their concrete
+types. With both backends enabled, publishing uses both concurrently;
+`NetworkOnly` resolution waits for both and selects the most recent result,
+while `CacheFirst` may finish as soon as a fresh result above the cache floor is
+available.
```
SeverinAlexB: I am really not convinced yet that these internal implementation details need to go in this README. This README is shown on crates.io so this is the main entrypoint for a lot of devs. At the moment, this is way too complicated for first time users.

Here is what AI says:

A crates.io README should help a developer answer four questions quickly:
1. What does this crate do?
2. How do I install it?
3. What is the smallest working example?
4. Which features do I need?

The current pkarr/README.md begins with architecture before showing installation or usage. That material is useful, but it should come after the quick start.

pkarr/README.md
```diff
@@ -1,23 +1,59 @@
...
+for configuration examples and the
+[API documentation](https://docs.rs/pkarr/latest/pkarr/) for the full public
+surface.
+
+## Runtime and Platform Support
+
+- The client API is asynchronous.
+- On native targets, `async_compat` allows use with non-Tokio executors.
+- Browsers use the `relays` feature and Fetch-backed HTTP requests because they
+  cannot access the UDP DHT directly.
```
SeverinAlexB: Can we simplify this sentence. It's very hard to ready

relay/README.md
```diff
@@ -1,28 +1,50 @@
 # Pkarr Relay
 
-A server that functions as a [pkarr relay](https://github.com/pubky/pkarr/blob/main/design/relays.md).
+The `pkarr-relay` binary is an HTTP gateway and cache for publishing and
+resolving Pkarr packets through the Mainline DHT. It implements the
+[Pkarr relay protocol](../design/relays.md).
```
SeverinAlexB: Do we have a "relay protocol"?

relay/README.md
```diff
@@ -1,28 +1,50 @@
...
 
+```bash
+pkarr-relay --tracing-env-filter pkarr_relay=debug,tower_http=debug
+```
+
+For local development, start an isolated DHT and relay on port `15411`:
+
+```bash
+pkarr-relay --testnet
+```
```
SeverinAlexB: Looking at this as a Synoym outsider must be weird. What does testnet mean in this case? Can I supply DHT bootstrap nodes or what is a pkarr testnet?

I assume this was an undocumented feature for development so far. If this is going to be made public, we should improve the feature first.

README.md
```diff
@@ -78,6 +89,12 @@ sequenceDiagram
     Relay->>Client: Verified response
 ```
 
+The default native client configures both paths. `ClientBuilder` can instead
+select DHT-only or relay-only operation, and `ResolvePolicy` controls whether a
+lookup uses cached data or queries the configured networks. See the
+[integration guide](./docs/integration.md#client-abstractions) for the client
+structure and selection behavior.
```
SeverinAlexB:
```suggestion
The default native client configures both paths. `ClientBuilder` can
select DHT-only or relay-only operation, and `ResolvePolicy` controls whether a
lookup uses cached data or queries the configured networks. See the
[integration guide](./docs/integration.md#client-abstractions) for the client
structure and selection behavior.
```

SeverinAlexB (changes requested):
Please AI review the changes for grammar mistakes. There are a lot of missing articles and other issues.

docs/integration.md
andrei-21: They are not the same, but I agree we should use `backend` in such cases. Fixed.

docs/integration.md
andrei-21: Agree, they are too detailed. Removed.

pkarr/README.md
andrei-21: Fair. Removed and simplified.

pkarr/README.md
andrei-21: I refactored it.

relay/README.md
andrei-21: Not that defined. I changed wordings.

relay/README.md
andrei-21: Ok, I just removed it.

README.md
andrei-21: Fixed.

README.md
```diff
@@ -52,6 +54,17 @@ async fn main() -> anyhow::Result<()> {
 | **[Examples](./pkarr/examples/README.md)** | Code samples |
 | **[Specifications](./design/README.md)** | Protocol design documents |
 
+## Repository Components
+
+- **`pkarr`** — The Rust library for keys, signed DNS packets, caching,
+  publishing, and resolution.
+- **`pkarr-relay`** — An HTTP-to-DHT relay server for browsers and other
+  UDP-restricted clients. Install it with `cargo install pkarr-relay`, or run
```
86667: Install instructions here is too much. A description and pointer to its own docs is fine. Also perhaps a line on why this crate is useful

pkarr/README.md
```diff
@@ -1,23 +1,71 @@
...
+```
+
+The client API is asynchronous. Tokio is used here to run the example; on
+native targets, Pkarr also supports other executors through `async_compat`.
+
+## Choosing Features
+
+| Use case | Dependency |
+|----------|------------|
+| Native application using DHT and relays | `pkarr = "7"` |
```
86667: Should be version 8, or replace with placholder

pkarr/README.md
```diff
@@ -1,23 +1,71 @@
...
-- **Limitations**: 
-  - Not compatible with WASI
-  - Cannot use WASI bindings for direct DHT access
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let keypair = Keypair::random();
+    println!("Public key: {}", keypair.public_key());
+
+    let packet = SignedPacket::builder()
+        .txt("_hello".try_into()?, "world".try_into()?, 300)
```
86667, line 30: Two quick start examples here and in pkarr/README which are the same. May be better to keep the root readme as an intro explanation and point to crate readmes for instructions.

86667: there are also subtle differences between the two quick start examples

86667 (changes requested):

pkarr/README.md
andrei-21: Removed the root readme example and keep only this one.

pkarr/README.md
andrei-21: Replaced with `cargo add pkarr` to avoid bumping the version very time we release a new major version.

README.md
andrei-21: Fair, removed and referenced the corresponding readme.

86667 (approved):

PR #287, opened 2026-08-05
refactor(dht)!: decouple public API from mainline

andrei-21:
Add pkarr-owned DHT configuration and request filtering, keeping
Mainline types and packet conversions internal to the DHT adapter.

BREAKING CHANGE: `pkarr::mainline`, Mainline-based `DhtClient::build`,
and SignedPacket/MutableItem conversions are removed. Configure DHT
clients and relays with `pkarr::dht::DhtConfig`.

### DIFF
--- a/pkarr/src/dht/config.rs
+++ b/pkarr/src/dht/config.rs
@@ -1,0 +1,107 @@
+use std::{
+    fmt::Debug,
+    net::{Ipv4Addr, SocketAddrV4},
+    sync::Arc,
+    time::Duration,
+};
+
+/// Filters incoming DHT requests.
+pub trait RequestFilter: Debug + Send + Sync {
+    /// Return whether a request from `from` should be handled.
+    fn allow_request(&self, from: SocketAddrV4) -> bool;
+}
+
+/// Configuration for a [`super::DhtClient`].
+#[derive(Clone, Debug)]
+pub struct DhtConfig {
+    /// UDP port used by the DHT node.
+    ///
+    /// When `None`, the DHT implementation's default port is used.
+    pub port: Option<u16>,
+    /// Whether the DHT node should answer incoming requests.
+    pub server_mode: bool,
+    /// DHT nodes used to bootstrap discovery.
+    ///
+    /// When `None`, the DHT implementation's default bootstrap nodes are used.
+    pub bootstrap: Option<Vec<SocketAddrV4>>,
+    /// Local IPv4 address to bind the DHT socket to.
+    ///
+    /// When `None`, the operating system chooses the local address.
+    pub bind_address: Option<Ipv4Addr>,
+    /// Maximum duration for a DHT request.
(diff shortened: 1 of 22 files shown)


86667 (commented):
Nice. A couple

pkarr/src/dht/config.rs
```diff
@@ -0,0 +1,107 @@
...
+
+/// Filters incoming DHT requests.
+pub trait RequestFilter: Debug + Send + Sync {
+    /// Return whether a request from `from` should be handled.
+    fn allow_request(&self, from: SocketAddrV4) -> bool;
+}
+
+/// Configuration for a [`super::DhtClient`].
+#[derive(Clone, Debug)]
+pub struct DhtConfig {
```
86667, line 17: `mainline::Config` has a `public_ip: Option<Ipv4Addr>` field for BEP_0042 secure node IDs. We remove that functionality here, a relay node may want to use BEP_0042 though i think?

pkarr/src/dht/config.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub(super) fn into_mainline(self) -> mainline::Config {
+        let mut config = mainline::Config {
+            bootstrap: self.bootstrap,
+            bind_address: self.bind_address,
+            request_timeout: self.request_timeout,
+            server_mode: self.server_mode,
+            ..Default::default()
+        };
+
+        if let Some(port) = self.port {
```
86667: nit: Both `port` types are the same. No need for conditional

relay/Cargo.toml
```diff
@@ -21,6 +21,7 @@ categories = ["network-programming"]
 [dependencies]
 anyhow = "1"
 axum = "0.8"
+mainline = { workspace = true }
```
86667: is it perhaps a sign that the abstraction isnt quite correct if this `mainline` import is needed here?

86667: I see that relay is using `mainline::Config` still. Can it use `DhtConfig`?

86667: And then if its only the test code that uses mainline directly then lets move into `[dev-dependencies]`

86667 (changes requested):
Nice, looks good. 

A couple comments from AI review. Main point is the added import to `relay`

pkarr/src/dht/config.rs
andrei-21: Fair, added.

pkarr/src/dht/config.rs
andrei-21: Nice.

relay/Cargo.toml
andrei-21: In practice `mainline` feature is used only for tests, so I hid it behind a new `testnet` feature.

86667 (approved):

PR #288, opened 2026-08-06
chore: release v8.0.0

andrei-21:
### Breaking changes

  - Gate relay testnet support behind explicit configuration.
  - Upgrade `simple-dns` to 0.12.
  - Upgrade `mainline` to v8.

  ### Added

  - Add secure node IDs to the relay.
  - Add DHT configuration support required by the updated networking stack.

  ### Changed

  - Update Rust workspace dependencies.
  - Update Dependabot scheduling and configuration.

### DIFF
--- a/relay/Cargo.toml
+++ b/relay/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pkarr-relay"
-version = "1.0.0"
+version = "2.0.0"
 rust-version = "1.85"
 authors = [
     "Nuh <nuh@nuh.dev>",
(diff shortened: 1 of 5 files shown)


relay/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pkarr-relay"
-version = "1.0.0"
+version = "2.0.0"
```
86667, line 3: Shall we keep pkarr and pkarr-relay's versions in-step, like we do in pubky-homeserver?

86667 (changes requested):

relay/Cargo.toml
andrei-21: Not sure. A breaking change in `pkarr-relay` will always result in a major version bump for `pkarr`, but it is not necessary what we want.

86667 (approved):

PR #298, opened 2026-08-31
chore: release v8.0.1

andrei-21:
### Added

- Configurable relay connection count and lifetime limits.
- HTTP connection, header, body, and drain timeouts.

### Changed

- Hardened HTTP/1 and HTTP/2 handling against resource exhaustion.
- Replaced `axum-server` with direct Hyper/Tokio handling.
- Reduced relay error-response memory usage.
- Improved client, relay, configuration, and feature documentation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1701,7 +1701,7 @@ checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 
 [[package]]
 name = "pkarr"
-version = "8.0.0"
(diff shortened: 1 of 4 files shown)


86667 (approved):
Approved, but i think theres a good argument for relay being a minor bump. wdyt?

andrei-21:
> Approved, but i think theres a good argument for relay being a minor bump. wdyt?

I set it to a patch because the main thing is the fix. The new config options (which would make it a minor) are not really advertised.

PR #306, opened 2026-09-08
fix: preserve endpoint resolution errors and alternatives

ben-kaufman:
## Summary

- Add fallible endpoint discovery so callers can distinguish lookup failure from an empty endpoint list.
- Preserve initial and alias lookup errors, and continue to remaining sibling endpoints when an alias fails.
- Keep operational errors when no usable packet was found and another backend returned `NotFound`. A valid packet still wins over an ordinary backend failure, and all-empty results still return `NotFound`.

## Why

A transient PKARR failure can disappear inside endpoint discovery. The Pubky SDK can then infer and cache a direct-transport choice from an empty result. Separately, mixed `NotFound + lookup failure` results can be misclassified as record absence.

These behaviors were reproduced with isolated loopback servers and real signed PKARR packets. The dependent [Pubky SDK PR #598](https://github.com/pubky/pubky-homeserver/pull/598) stops caching failed transport discovery.

## Review scope

The reqwest connector and socket-address conversion changes have been removed. Those files now match the base revision. This PR no longer adds candidate-address collection, an asynchronous DNS helper, or a 100 ms backup-discovery budget.

The remaining endpoint changes expose errors through the existing discovery API and preserve valid sibling alternatives. Broader endpoint/request ownership and connector failover are outside this PR.

The mixed negative/error-result precedence is an intentional behavior change and is still under review.

## Staging logs still needed

The team reports that Locks was rate-limited by the PKARR relays and that the DHT's `NotFound` masked the relay failure. This PR addresses error handling, not the rate limit itself. Disabling DHT avoids that particular mixed-backend path but does not remove relay rate limiting or the SDK's failed-discovery cache bug.

We still need a correlated failing auth attempt to confirm the exact errors, the rate-limit policy being hit, and whether the session POST reached the homeserver:

- UTC timestamp/request ID and deployed versions.
- Locks-side SDK/PKARR logs, including the full underlying error chain and relay/DHT lookup results, not just the final 503.
- Matching relay logs with response status and rate-limit details, and homeserver logs where available.

Please redact credentials, cookies, auth tokens, and request bodies, and share sensitive logs privately. The local reproduction does not establish every step of the historical staging failures.

## Validation

- 209 PKARR library tests passed, including retained regression cases.
- Strict Clippy for the library and tests passed.
- `cargo fmt --all -- --check` and diff-whitespace checks passed.
- WASM check with `relays,endpoints` passed.
- Loopback checks passed for failed-alias sibling continuation and mixed relay results under both `CacheFirst` and `NetworkOnly`.
- A valid packet plus a failing relay still succeeds. All-`NotFound` remains `NotFound`.
- The dependent SDK's 201 library tests and real loopback recovery diagnostic passed against the narrowed source.

Existing endpoint APIs are retained. No application-level HTTP retries, authentication/session POST replay, or TLS-probing redesign are added.

### DIFF
--- a/pkarr/src/client/backend/resolve_result_accumulator.rs
+++ b/pkarr/src/client/backend/resolve_result_accumulator.rs
@@ -100,8 +100,9 @@ impl<'a> ResolveResultAccumulator<'a> {
     /// the selected valid packet. With cache-first constraints, an expired
     /// packet may be returned when no fresh candidate exists so the client can
     /// update its cache before returning [`ResolveError::NotFound`] to its caller.
-    /// If every usable response was empty or covered by the floor, returns
-    /// [`ResolveError::NotFound`].
+    /// Without a packet, operational errors take precedence over empty responses.
+    /// Returns [`ResolveError::NotFound`] only when all responses were empty or
+    /// covered by the floor and no backend failed.
     pub(super) fn into_result(self) -> Result<SignedPacket, ResolveError> {
         let packet = self.fresh_candidate.or(self.most_recent);
 
@@ -110,7 +111,7 @@ impl<'a> ResolveResultAccumulator<'a> {
             (Some(seq), Some(packet)) if packet.timestamp().as_u64() as i64 >= seq => Ok(packet),
             (Some(seq), _) => Err(ResolveError::InvalidSignedPacket { seq }),
             (None, None) => {
-                if self.empty_responses > 0 {
+                if self.empty_responses > 0 && self.errors.is_empty() {
                     Err(ResolveError::NotFound)
                 } else {
                     Err(most_common_error(self.errors))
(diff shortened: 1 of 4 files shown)


pkarr/src/client/backend/resolve_result_accumulator.rs
```diff
@@ -110,7 +111,7 @@ impl<'a> ResolveResultAccumulator<'a> {
             (Some(seq), Some(packet)) if packet.timestamp().as_u64() as i64 >= seq => Ok(packet),
             (Some(seq), _) => Err(ResolveError::InvalidSignedPacket { seq }),
             (None, None) => {
-                if self.empty_responses > 0 {
+                if self.empty_responses > 0 && self.errors.is_empty() {
```
andrei-21, line 114: I see the issue with the logic, but I doubt it is the fix. In case of one faulty relay we will fail the entire call, we should not do it.

pkarr/src/extra/reqwest.rs
andrei-21, line 1: All endpoints and request logic is weird, I am not sure it should belong here, so I doubt we should improve it.

andrei-21 (changes requested):

pkarr/src/client/backend/resolve_result_accumulator.rs
ben-kaufman: A valid packet still wins when another relay fails, and the tests cover that. This only changes the case where no packet was found and one backend returned `NotFound` while another failed. I kept the lookup error there so we don't report a failed lookup as record absence. Are you referring to that mixed `NotFound` + failure case?

On your `no_dht()` suggestion: thanks, makes sense to try it to avoid the misleading `NotFound`. But if the relays are rate limiting us, wouldn't auth still fail, just with a different error? Do we know which limit we're hitting and why? The SDK cache fix should still help avoid keeping that failure around after the lookup recovers.

pkarr/src/extra/reqwest.rs
ben-kaufman: Agreed, I removed the reqwest/address-fallback changes and timeout policy in b09864db53bf3f4a95e877c2c9cc01a7b8a364af. Those files now match the base revision. It keeps error reporting through the existing endpoint discovery API, including continuing past failed aliases, so #598 can avoid caching failed discovery. We can discuss where endpoint/request handling belongs separately.

pkarr/src/client/backend/resolve_result_accumulator.rs
dzdidi: DISCLAIMER: I do lack context and understanding of the codebase. But I agree that it is hard to judge if this is a correct fix or not, but I have a suspicion why this is the case.

The code collects responses into different properties of `self`  like `fresh_candidate`, `most_recent`, `highest_invalid_signed_packet_seq`, `empty_responses`, `errors`  and then based on their combination code tries to decide what is the result. This is basically complex state on one dimension and complex logic on another dimension.

I think having single queue of responses (simple state) which then gets filtered by response type, filtered by sequence, reduced into result etc  will make it a) easier to reason b) easier to change if needed.

